PR #22, opened 2016-12-13
Display door flags on door names [DO NOT MERGE]

teoli2003:
Adds a [h/NL] or [d] (and various combination) to door name with these flags.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,30 +1,37 @@
-=============================
-MMapper 2.3.6 December 9, 2015
-=============================
+================================
+MMapper 2.3.7 (Not released)
+================================
+
+Changes:
+  - Display some door flags next to door names (teoli)
+
+================================
+MMapper 2.3.6 (December 9, 2015)
+================================
 
 Changes:
   - High DPI displays are now supported such as Retina displays (nschimme)
 
-=============================
-MMapper 2.3.5 July 29, 2015)
-=============================
+=================================
+MMapper 2.3.5 (July 29, 2015)
+=================================
 
 Changes:
   - Fixed bug that prevented connections from having TCP KeepAlive (nschimme)
   - Updated base map to include new zones; thanks Ortansia! (nschimme)
 
-=============================
+================================
 MMapper 2.3.4 (May 1, 2015)
-=============================
+================================
 
 Changes:
   - All connections now utilize TCP KeepAlive to help with dropped connections (nschimme)
   - Prompts are now correctly identified and remembered for internal commands (nschimme)
   - Cleaned up vote and Windows build code (nschimme)
 
-=============================
+================================
 MMapper 2.3.3 (January 18, 2015)
-=============================
+================================
 
 Changes:
   - [GroupManager] Player's hp, mana, and moves are now correctly updated (nschimme)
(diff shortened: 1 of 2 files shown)


nschimme:
I'm a little concerned about breaking user scripts that rely on the previous behavior. Maybe we should include different lock icons or color codes rather than doing this?

NEWS.txt
```diff
@@ -1,30 +1,37 @@
...
+================================
 
 Changes:
   - High DPI displays are now supported such as Retina displays (nschimme)
 
-=============================
-MMapper 2.3.5 July 29, 2015)
-=============================
+=================================
+MMapper 2.3.5 (July 29, 2015)
```
nschimme: Thanks for fixing this 👍

src/display/mapcanvas.cpp
```diff
@@ -1490,6 +1490,50 @@ void MapCanvas::drawRoomDoorName(const Room *sourceRoom, uint sourceDir, const R
...
+    {
+      targetPostFix = "h";
+    }
+    if (ISSET(targetDoorFlags, DF_NEEDKEY))
+    {
+      targetPostFix += "L";
+    }
+    if (ISSET(targetDoorFlags, DF_NOPICK))
+    {
+      targetPostFix += "/NP";
```
nschimme: Why add a "/" in front of this combination and not others?

nschimme:
I thought about this some more and couldn't find a clever way of distinguishing between the types on the UI. The main problem is up/down doors where the usable UI is extremely constrained. This really only left color coding as a possible solution on either the text or "door cross" but since everything is a combination of 4 states I'm not entirely sure if that would work.

Your current solution probably makes the most sense unless you can think of something else.

src/display/mapcanvas.cpp
```diff
@@ -1490,6 +1490,50 @@ void MapCanvas::drawRoomDoorName(const Room *sourceRoom, uint sourceDir, const R
...
+    }
+    DoorFlags targetDoorFlags = getDoorFlags(targetRoom->exit(targetDir));
+    QString targetPostFix;
+    if (ISSET(targetDoorFlags, DF_HIDDEN))
+    {
+      targetPostFix = "h";
+    }
+    if (ISSET(targetDoorFlags, DF_NEEDKEY))
+    {
+      targetPostFix += "L";
```
nschimme: Why "L" for "lock" rather than "K" for "key"?

src/display/mapcanvas.cpp
```diff
@@ -1490,6 +1490,50 @@ void MapCanvas::drawRoomDoorName(const Room *sourceRoom, uint sourceDir, const R
     qint32 dX = srcX - tarX;
     qint32 dY = srcY - tarY;
 
+    // Look at door flags to code postfixes
+    DoorFlags sourceDoorFlags = getDoorFlags(sourceRoom->exit(sourceDir));
```
nschimme: Do we not care about the other two types of DF_NOBLOCK and DF_NOBREAK ?

teoli2003:
I'm putting this PR on hold. I would like to think more about this and there is no impact on the file schema so this can wait.

nschimme:
Any chance you can get an update to this over the winter break? I'm planning on doing a release soon(tm)

src/display/mapcanvas.cpp
nschimme: I started thinking about this again. Maybe instead of using /NP we can add ANSI color codes to the "L" in the lock? Perhaps red? :-)

src/display/mapcanvas.cpp
nschimme: How often are these flags used independently? Can we use one flag for NOBLOCK and NOBREAK?

nschimme:
@teoli2003 any updates?

nschimme:
Looks like this got merged outside of the NEWS.txt :-)

PR #23, opened 2016-12-15
Fix wall color when special exit is present

teoli2003:
When a special exit is present, 2.3.6 (and maybe others) set the default color for walls to be drawn after the special exit to 'white'.

It should have been 'black'. 

This bug fix it. I wasn't able to trace back where the bug was introduced, it may have been a side effect of another change, likely when High DPI screen support was added.

(There still is another bug where the stipple colors are lost, but it is independent, and I have not yet figured it out)

I have based this PR on top of PR#22 (there only dependency is the bump in version). I can rebase if needed.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,6 +1,12 @@
-=============================
-MMapper 2.3.6 December 9, 2015
-=============================
+===============================
+MMapper 2.3.7 (Not published)
+===============================
+Changes:  
+  - Fixed wall color when a special exit was present (was white instead of black) (teoli)
+
+================================
+MMapper 2.3.6 (December 9, 2015)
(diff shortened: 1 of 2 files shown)


nschimme:
In the past I have waited to bump versions whenever I had the time to do a QA test on Mac/Linux/Windows.

nschimme:
LGTM. Will merge if this wasn't on top of the doors PR.

teoli2003:
I will rebase, removing the version change (proved to be a bad idea) and directly over 3.2.6 :-)

teoli2003:
Rebased done for this one.

nschimme:
👍

PR #24, opened 2016-12-16
Ability to have classes of infomarks and to rotate them

teoli2003:
I've extended the infomarks feature.

Note: this needed an extension of the schema (now at version 040) and I therefore bumped the version number to 2.4.0 as old map are compatible, but map files saved with this version can't be opened in 2.3.x and earlier version (semantic versioning).

Info marks have a class: there are several of them, like an info mark of class 'locality' to describe a village name, or a 'road', a 'river', an 'action', …

Most classes have a different rendering: a river info mark will be displayed in blue, a road in brown, … Localities are underlined and actions are in italics.

This leads to much more neat maps. The color schemes allow to add more infomarks without having them being too visible. This is especially true for roads and rivers than blends a bit with the tile, allowing to add more of these. 

Another cool feature is the ability to rotate info marks. River or road names benefit of it as they can follow the direction of them.

I've added an example screenshot of the result on maps:
screen shot 2016-12-16 at 09 08 19


src/display/mapcanvas.cpp
```diff
@@ -2970,19 +3054,30 @@ void MapCanvas::qglClearColor(QColor clearColor) {
     glClearColor(clearColor.redF(), clearColor.greenF(), clearColor.blueF(), clearColor.alphaF());
 }
 
-void MapCanvas::renderText(double x, double y, const QString &text) {
+void MapCanvas::renderText(double x, double y, const QString &text, QColor color, uint fontFormatFlag, double rotation) {
     // http://stackoverflow.com/questions/28216001/how-to-render-text-with-qopenglwidget/28517897
     int height = this->height();
 
     GLdouble textPosX = 0, textPosY = 0, textPosZ = 0;
-    Project(x, y, 1.2f, &textPosX, &textPosY, &textPosZ);
```
nschimme: If I remember correctly this was 1.2f in the old code when I ported it from Qt4 to Qt5. Not sure if it is still required. Are Z-layers behaving correctly with this set to 0.0f?

src/mapstorage/mapstorage.cpp
```diff
@@ -955,6 +975,8 @@ void MapStorage::saveMark(InfoMark * mark, QDataStream & stream)
   stream << (QString)mark->getText();
   stream << (QDateTime)mark->getTimeStamp();
   stream << (quint8)mark->getType();
+  stream << (quint8)mark->getClass();
+  stream << (qint32)(mark->getRotation()*100);
```
nschimme: If you agree that a better name is `m_rotationAngle` then this should be changed to `getRotationAngle()` as well.

src/display/mapcanvas.cpp
teoli2003: I removed this as the labels were floating over the map, which led to them always positioned a bit wrongly.

src/mapdata/infomark.h
```diff
@@ -32,39 +32,46 @@
...
     InfoMarkType m_type;
     InfoMarkText m_text;
-    
+    InfoMarkClass m_class;
+
     MarkerTimeStamp m_timeStamp;
 
     Coordinate m_pos1;
     Coordinate m_pos2;
+    double m_rotation; // in degrees
```
teoli2003: Will fix.

src/mapstorage/mapstorage.cpp
teoli2003: Yep :-)

src/mapdata/infomark.h
teoli2003: Fixed

src/mapstorage/mapstorage.cpp
teoli2003: Fixed

teoli2003:
So I rebased it not to have the dependencies on the PR for the door names, and I fixed the review comments.

nschimme:
👍

PR #26, opened 2016-12-17
Resurrect qmake .pro build script and include zlib as a 3rdparty library

nschimme:
Mac OS X has broken CMake with the Sierra upgrade. I've resurrected the older build system even though it doesn't provide packaging since it has better support.

@teoli2003 let me know if this works for you too


teoli2003:
This is strange. I'm compiling (successfully) with CMake 3.7.1 and Qt 5.7 on macOS Sierra 10.12.1.

nschimme:
Ah. Hmm, maybe because I used `brew` to install everything.

nschimme:
Found the problem. I had to either do `export CMAKE_PREFIX_PATH /usr/local/opt/qt5/` or update `CMakeLists.txt` to do that automatically. I'll submit another PR to fix this.

nschimme:
Gah. I can't figure out why my environment is messed up. For some reason anything touched by `macdeployqt` ends up crashing when it tries to initialize OpenGL.

PR #33, opened 2017-01-11
Command line access to set and print note

ethorondil:
This adds two new commands: _note to set a note in the current room, and _pnote to print the current note.

Also makes _pdynamic and _pstatic work; they'd previously bitrotted to print nothing.


nschimme:
👍

PR #34, opened 2017-01-11
Shortest path directions

ethorondil:
Adds a new command, _dirs, that prints the shortest path to rooms satisfying a condition.

By default it searches the room name, but different flags can give different options.  The format is
`_dirs [-(name|desc|note|exits|all)] search pattern`.

For example,
```
>_dirs caras
Distance 19.45: The Gates of Caras Galadhon
dirs: e3n4e2ne
Distance 20.05: Before the Walls of Caras Galadhon
dirs: e3n2ene3n2e
>_dirs -a ferry
Distance 14.9: Landing in the Forest
dirs: 2en3ese
Distance 64.75: A Landing
dirs: 2s6e
Distance 586.45: Road to Standelf
dirs: nwn6ws3wn6w2s4wsw2sw3s8ws9wn4w11nwn2wu8wsw3de3dn8ws8wswsw2usds3wn3wnws3wu6wswnwnwnwnwn6ws5w2sw4s3wsws2ws3wsws4wsws3ws3wsws4ws2wswsws10ws2wn6wn2wn6wsws4wnwnwnwn3wnw2n9ws3w4s4w2s2ws2ws6wnwnwnwnwnw2n2e2n2w2ne10nwnwnwnwnwn2wnwnw2nwnwnwnwn3wnwn3wn7wn2wn2w2nwn4w3n2w2nw5nw2nw8nw3ne3ne2nene2ne4ne5nenene2ne4ne4ne4n2wnw6nw5n2w4n3wn2wn6wn5ws4wswsw4sws2ws
[...]
```
(Why does the path to the ferry jog north? Because doing so avoids a noride room.)

or you can search for a specific exit, e.g.:
```
>_dirs -e hatch
Distance 13.75: An Abandoned Cellar
dirs: wdse8sw2usw
Distance 16.25: Old Ruins
dirs: wdse8sw2uswu
[...]
```
or search your notes with `_dirs -n` or `_dirs -note`.

The last commit here adds a delay to offline character moving, giving it time to sync up when speedwalking.

[This is a pretty big commit, so you probably want me to reorganize it/conform to the style more; just let me know how.  I also have a similar _search command to highlight rooms, which shares the filter code here and for which I'll submit a PR once this lands.]

### DIFF
--- a/src/mapdata/roomfilter.cpp
+++ b/src/mapdata/roomfilter.cpp
@@ -2,7 +2,7 @@
 #include <assert.h>
 #include <errno.h>
 
-const char * RoomFilter::parse_help = "Parse error; format is: [-(name|desc|note|exits|all)] text\r\n";
+const char * RoomFilter::parse_help = "Parse error; format is: [-(name|desc|note|exits|all)] pattern\r\n";
 
 
 bool RoomFilter::parseRoomFilter(const QString & line, RoomFilter &output)
--- a/src/mapdata/roomfilter.cpp
+++ b/src/mapdata/roomfilter.cpp
@@ -1,0 +1,54 @@
+#include "roomfilter.h"
+#include <assert.h>
+#include <errno.h>
+
+const char * RoomFilter::parse_help = "Parse error; format is: [-(name|desc|note|exits|all)] text\r\n";
+
+
+bool RoomFilter::parseRoomFilter(const QString & line, RoomFilter &output)
+{
+  QString pattern = line;
+  char kind = PAT_NAME;
+  if (line.size() >= 2 && line[0] == '-')
+  {
+    QString kindstr = line.section(" ", 0, 0);
+    if (kindstr == "-desc" || kindstr == "-d") kind = PAT_DESC;
+    else if (kindstr == "-name") kind = PAT_NAME;
+    else if (kindstr == "-exits" || kindstr == "-e") kind = PAT_EXITS;
+    else if (kindstr == "-note" || kindstr == "-n") kind = PAT_NOTE;
+    else if (kindstr == "-all" || kindstr == "-a") kind = PAT_ALL;
+    else return false;
+
+    pattern = line.section(" ", 1);
+  }
+  output = RoomFilter(pattern, Qt::CaseInsensitive, kind);
+  return true;
+}
+
+const bool RoomFilter::filter(const Room *r) const
+{
+  if (kind == PAT_DESC)
+    return QString((*r)[1].toString()).contains(pattern, cs);
+  else if (kind == PAT_ALL)
+  {
+    for(auto elt: (*r))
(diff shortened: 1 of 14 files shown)


src/parser/abstractparser.cpp
```diff
@@ -856,6 +925,21 @@ bool AbstractParser::parseUserCommands(QString& command)
     else
       if (str.startsWith("_block"))   {dooraction = true; daction = DAT_BLOCK;}
 
+    if (str.startsWith("_dirs"))
+    {
+      QString pattern_str = str.section(' ', 1).trimmed();
+      if(pattern_str.size() == 0)
+      {
+        emit sendToUser("Usage: _search <pattern in quotes> [section to search]\r\n");
```
nschimme: Should be _dirs rather than _search ?

src/parser/abstractparser.cpp
```diff
@@ -856,6 +925,21 @@ bool AbstractParser::parseUserCommands(QString& command)
     else
       if (str.startsWith("_block"))   {dooraction = true; daction = DAT_BLOCK;}
 
+    if (str.startsWith("_dirs"))
```
nschimme: The _dirs command and usage should also be added to the _help prompt

src/parser/abstractparser.cpp
```diff
@@ -1201,6 +1285,7 @@ void AbstractParser::offlineCharacterMove(CommandIdType direction)
     }
   }
   m_mapData->unselect(rs1);
+  usleep(40000);
```
nschimme: Any reason why we needed this usleep?

src/mapdata/roomfilter.cpp
```diff
@@ -0,0 +1,54 @@
+#include "roomfilter.h"
+#include <assert.h>
+#include <errno.h>
+
+const char * RoomFilter::parse_help = "Parse error; format is: [-(name|desc|note|exits|all)] text\r\n";
```
nschimme: This should probably also be documented somehow. Or maybe just mention _dirs  -help ?

src/parser/abstractparser.cpp
ethorondil: This is a hackish fix to the following issue: in trunk, speedwalking doesn't work offline.  This makes it work (at least for me, at my computer's speed).

I don't really understand the root of the issue.  This makes it a little bit slow for long journeys, but at the same time it lets you watch the path taken, so it's not too bad.

ethorondil:
OK, I've fixed the easy documentation issues.  Since it's hard to explain everything in one line, it might be good to have a place for more extensive documentation, like `_help _dirs` or something.

src/parser/abstractparser.cpp
nschimme: Seems fine to me since it's only limited to offline character moves.

nschimme:
:+1:

Issue #5, opened 2015-02-03
Add a new command _search to search for notes


nschimme:
it would be nice to have console interface for searching (including the new note-search), e.g. comand _search

PR #35, opened 2017-02-06
Search function

ethorondil:
Adds a `_search` function similar `_dirs`, except it highlights matching rooms rather than prints directions to them.  I'm not quite sure why selected things are highlighted in red -- maybe it was intended to debug leaks? -- but it seems to work ok.

Addresses #5 (just `_search -n note text`).


ethorondil:
Probably a lot of map editing commands should use the `emit m_mapData->updateCanvas();` signal introduced in this PR, so the map is updated immediately when room flags are toggled.

nschimme:
👍

PR #36, opened 2017-02-07
Fix exits/emulated behavior

ethorondil:
Currently, the Exits/emulated: lines are pretty buggy, and often
display information for the wrong room.  The output is printed before
the character's location is synced, so mmapper tries to guess where
the room will be based on the queued movement commands.
Unfortunately, this is pretty problematic:

 * following doesn't work, because the order MUME sends information is
    like:

Arwen Undómiel leaves east.
You follow Arwen Undómiel.
<movement dir=east/><room><name>The Courtyard</name>

m_move is set to CID_EAST on the <movement> tag, but it's enqueued the
line before, on the "You follow" line, at which point it's still
CID_LOOK.

 * Scouting doesn't work, because the queue is prepended with
   CID_SCOUT rather than CID_EAST.

 * Regular movement in the Old Forest doesn't work, because the
   prediction isn't reliable -- even if you go to a recognizable room.

 * Teleporting/scrying don't work, because there's no way to guess the
   location from just the commands.

Instead, the emulated exits should be printed _after_ the pathmachine
syncs up.

This patch makes that happen in XML mode.  It doesn't fix it for
regular mode, but it shouldn't make things worse there either.

It doesn't fix teleport/scry, though.  Scry is hard, since you don't really want to move the character to where you're scrying.  Teleport should be doable in principle, but it's a bit tricky.  Pathmachine doesn't go from ACCEPTED to SYNCING in one try, and I don't know how the parser can identify teleportation to tell pathmachine to releaseAllPaths.  Maybe when you receive a <movement/> tag, it should prioritize nearby rooms (e.g. because of currents) but make sure to sync globally if that doesn't work?

### DIFF
--- a/src/expandoracommon/parseevent.h
+++ b/src/expandoracommon/parseevent.h
@@ -29,6 +29,7 @@
 #include <deque>
 #include "listcycler.h"
 #include <QVariant>
+#include <QMutex>
 
 class Property;
 
@@ -38,7 +39,7 @@ class Property;
 class ParseEvent : public ListCycler<Property *, std::deque<Property *> >
 {
 public:
-  ParseEvent(uint move) : moveType(move), numSkipped(0) {}
+  ParseEvent(uint move) : moveType(move), numSkipped(0) {parsedMutex.lock();}
   ParseEvent(const ParseEvent & other);
   virtual ~ParseEvent();
   ParseEvent & operator=(const ParseEvent & other);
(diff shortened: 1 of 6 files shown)


src/expandoracommon/parseevent.h
```diff
@@ -49,12 +50,16 @@ class ParseEvent : public ListCycler<Property *, std::deque<Property *> >
   const std::deque<QVariant> & getOptional() const {return optional;}
   uint getMoveType() const {return moveType;}
   uint getNumSkipped() const {return numSkipped;}
-  
+
+  void setParsed();
+  bool waitForParsed(int timeout=-1);
```
nschimme: Wouldn't it be safer to set this to some more reasonable value? What if the server freezes? Won't this lock up the entire thread?

ethorondil:
I don't think it's the direction for server freezes to imply deadlock, but you're probably right that I should set it to 10ms or something.

There are some bugs with the current implementation, though, so I'll wait a bit to make a new PR.

nschimme:
Any chance you can get an update to this over the winter break? I'm planning on doing a release soon(tm)

nschimme:
@ethorondil any updates?

PR #37, opened 2017-02-13
Fix occasional segfaults in _dirs

ethorondil:
The old code stored a pointer into the QVector sp_nodes, which could be invalid after elements are added.  This could occasionally cause segmentation faults.


nschimme:
:+1:

PR #46, opened 2017-12-03
Export Web Map (v1 format)

waba4mume:
This is the MM2 part of the recent [webclient+map release](https://github.com/waba4mume/play-mume):
- Added an Export Web Map menu entry.
  The v1 format is JSON with a roomname/desc -> location index for lazy downloading.
- Killed some duplicated latinToAscii() code.
- Fixed an old encoding bug causing NUL bytes in the map data, see commit message for 15e5f36.

@nschimme: I'll leave this PR up for a bit in case you want to review the code, otherwise I guess I'll just merge it myself. I'm not sure how you want the collaboration on MM2 to work.


src/mainwindow/mainwindow.cpp
```diff
@@ -1133,20 +1166,35 @@ bool MainWindow::saveFile(const QString &fileName, bool baseMapOnly )
...
-  if (storage->canSave()) storage->saveData( baseMapOnly );
+  if (storage->canSave())
+  {
+    bool baseMapOnly = mode != SAVE_FULL;
+    if ( !storage->saveData( baseMapOnly ) )
+    {
+      getCurrentMapWindow()->getCanvas()->setEnabled(true);
+      progressDlg->hide();
+      disableActions(true);
+      return false;
```
nschimme: We aren't closing the saver here.

src/mainwindow/mainwindow.h
```diff
@@ -84,8 +84,9 @@ class MainWindow : public QMainWindow
   Mmapper2PathMachine *getPathMachine(){return m_pathMachine;};
   CGroup *getGroupManager() {return m_groupManager;};
 
+  enum SaveMode { SAVE_FULL, SAVE_BASEMAP, SAVE_WEB };
```
nschimme: We're overloading `SaveMode` with `SaveFormat`. There isn't a way to save a WebBaseMap now. Not a major pain point for me but I did notice this.

src/mapstorage/jsonmapstorage.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+#include <cassert>
+#include <iostream>
+
+using namespace std;
+
+namespace
+{
+
+typedef QList<const Room *> RoomList2;
```
nschimme: Did you have a major reason to not re-use `RoomList`? I know it's a vector rather than a QList. Alternative we can also rename the current `RoomList` -> `RoomVector`.

src/mapstorage/jsonmapstorage.cpp
```diff
@@ -0,0 +1,589 @@
...
+  emit log ("JsonMapStorage", "Writing data finished.");
+
+  m_mapData.unsetDataChanged();
+  emit onDataSaved();
+
+  return true;
+}
+
+
+  /*
```
nschimme: Any reason to keep this comment?

src/mapstorage/jsonmapstorage.h
```diff
@@ -0,0 +1,63 @@
...
+
+#ifndef JSONMAPSTORAGE_H
+#define JSONMAPSTORAGE_H
+
+#include "abstractmapstorage.h"
+
+class QJsonArray;
+class QJsonObject;
+
+/*! \brief JSON export for web clients.
```
nschimme: Is there a schema you can link somewhere?

src/mainwindow/mainwindow.cpp
nschimme: Can we refactor this code to be simpler? I don't like the fact that the branching logic has semi-duplicate code regarding the UI.

src/configuration/configuration.cpp
```diff
@@ -258,6 +262,8 @@ void Configuration::write() const {
 
   conf.setValue("Static room description parser type", (int)m_roomDescriptionsParserType);
   conf.setValue("Minimum static lines in room description", m_minimumStaticLines);
+
+  conf.setValue("MUME charset is UTF-8", m_utf8Charset);
```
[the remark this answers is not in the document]
waba4mume: Setting encoding and color settings with MPI would be awesome, also because it would kill two prompts for new players with compatible clients.

This may be more complicated though, as both these settings have to match what the client behind MMapper expects. So the client would have to negociate it and MMapper would have to snoop that exchange and adapt its settings...

src/mainwindow/mainwindow.h
waba4mume: `SAVE_WEB` is actually (always) WebBaseMap. This was good enough for the proof of concept I was writing, but indeed at some point someone might want to make a full WebMap and improve Play MUME! to optionally point to it.

I would have also liked a customized save dialog where that can be set instead of that many menu entries, but I think that was too much work at the time.

src/mapstorage/jsonmapstorage.cpp
waba4mume: I think I wanted QList behavior for some reason. I'll review it and see if it could use that QVector type instead.

src/mapstorage/jsonmapstorage.cpp
waba4mume: Looks like unfinished work. It definitively won't make it into v1 format, I'll remove it.

src/mapstorage/jsonmapstorage.h
waba4mume: No, but that would be an interesting improvement if a future version of Play MUME! can efficiently make use of it.

waba4mume:
This code dates from the "let's see if that crazy idea can actually work" proof-of-concept era, and... it shows. I didn't go through it again before submitting this PR, I should have.

I'll come back with something cleaner.

src/mainwindow/mainwindow.cpp
waba4mume: The saver does not need explicit closing, its destructor handles that.

I refactored the logic in a separate `json4` branch, coming soon to a PR close to you...

src/mainwindow/mainwindow.cpp
```diff
@@ -1133,20 +1166,35 @@ bool MainWindow::saveFile(const QString &fileName, bool baseMapOnly )
...
+    }
+  }
   //getCurrentMapWindow()->getCanvas()->show();
   disableActions(false);
   cutAct->setEnabled(false);
   copyAct->setEnabled(false);
   pasteAct->setEnabled(false);
 
-  delete progressDlg;
```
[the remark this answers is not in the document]
waba4mume: This was a mistake, rolled back in `json4`: although the MainWindow's destructor would have eventually reaped progressDlg, it is more QTish to delete the dialog explicitely and let its destructor hide it.

Thank you for pointing this out.

src/mapstorage/jsonmapstorage.cpp
waba4mume: It can't be RoomList because in one instance it stores `const Room*`, not `Room *`, and in the other it needs to be compatible with `RoomSaver`.

In `json4` I clarify `RoomVector` (as you suggested) vs. `ConstRoomList`.

waba4mume:
Resubmitted as #48 (`json4` branch).

nschimme:
Great, thanks!

PR #48, opened 2017-12-05
Export Web Map (v1 format) - take 2

waba4mume:
Like #46 :

> This is the MM2 part of the recent [webclient+map release](https://github.com/waba4mume/play-mume):
> - Added an Export Web Map menu entry.
>   The v1 format is JSON with a roomname/desc -> location index for lazy downloading.
> - Killed some duplicated latinToAscii() code.
> - Fixed an old encoding bug causing NUL bytes in the map data, see commit message for a25864b.

Changes since #46 :
- Last 5 commits.
- (Rewritten history:) Roll back dialog delete vs. hide().
- (Rewritten history:) Remove commented-out code at the bottom of jsonmapstorage.cpp.


src/parser/mumexmlparser.cpp
```diff
@@ -338,8 +339,13 @@ QByteArray MumeXmlParser::characters(QByteArray& ch)
       m_lastPrompt = ch;
   }
 
-  m_stringBuffer = ch.simplified();
-  latinToAscii(m_stringBuffer);
+  if (Config().m_utf8Charset)
+      m_stringBuffer = QString::fromUtf8(ch);
```
nschimme: Do we need to convert this to Latin-1? We don't have an equivalent `ParserUtils::utf8ToAscii()`

src/parser/parser.cpp
```diff
@@ -112,7 +115,10 @@ void Parser::parseNewMudInput(IncomingData& data /*TelnetIncomingDataQueue& que*
 		(*debugStream) << "Prompt";
 		(*debugStream) << "***ETYPE***";
 #endif
-                m_stringBuffer = QString::fromLatin1(data.line.constData(), data.line.size());
+                if (Config().m_utf8Charset)
+                    m_stringBuffer = QString::fromUtf8(data.line);
```
nschimme: Same here regarding character encoding

src/parser/parserutils.cpp
```diff
@@ -0,0 +1,116 @@
...
+          'u',
+          'y',
+          'b',
+          'y'
+  };
+  unsigned char ch;
+  int pos;
+
+  for (pos = 0; pos < str.length(); pos++) {
+    ch = str.at(pos).toLatin1();
```
nschimme: Oh! You convert here to Latin1 :-)

nschimme: @waba4mume please confirm that UTF-8 is mapping to ASCII and we're good to merge :+1:

waba4mume:
```C++
if (Config().m_utf8Charset)
    m_stringBuffer = QString::fromUtf8(ch);
else
    m_stringBuffer = QString::fromLatin1(ch);
```

These lines make sure that the sequence of bytes that the MUME sends us is correctly decoded to QChars inside the QString (ie. `ë` becomes `ë`, not `Ã«` or an invalid sequence). QStrings store Unicode, so these `QString::fromXxx()` do not alter the logical content.

```C++
ParserUtils::latinToAscii(m_stringBuffer);
```

Then `ParserUtils::latinToAscii()` maps the QChars (Unicode chars) to their closest 7b ASCII equivalent. The author of that function called it latinToAscii() because it only handles chars in the Latin1 subset (= "chars that can be represented in the Latin1 encoding").

This works because MUME is also restricted to the Latin1 subset internally, so we cannot get a character that `latinToAscii()` cannot handle. If that restriction was lifted on MUME's side, these non-Latin1 chars would become NUL bytes, that's what `QChar::toLatin1()` returns for characters which cannot be represented as Latin1. NUL bytes are harmless in MMapper, although they break the web client's room lookup.

Expanding `latinToAscii()` to transliterate the whole Unicode range to ASCII is not feasible, we would have to pull in an extra library for that purpose, such as iconv or ICU.

There are no plans at the moment to refactor MUME's code to use Unicode strings everywhere, and even if someone was to undertake this huge task, it's unlikely that non-Latin1 chars end up in opened room names or descs.

So the current solution seems like a good compromise.

nschimme:
It's great that MUME is internally represented as Latin-1. I agree with your analysis.

Issue #17, opened 2015-08-23
Hinting the location of distant players


tomaswindsor:
It would be nice if MMapper hinted the direction of other connected players via group-manager in case their location is not rendered in currently visible part of the map.
Something like a small line or dot of player's colour would be rendered at border of the rendered map at appropriate position.

nschimme:
I've taken a stab at it and managed to learn some basic OpenGL in the process. Should be able to get in as part of 2.4.0 :-)

tomaswindsor:
Take a look at [game-play of old Atari 8-bit game Getaway](https://www.youtube.com/watch?v=GI7XGChVjd0) which implements that feature for distant sprites for inspiration.

nschimme:
Heh, much simpler than my solution of using an ellipse. Give the current solution a try and see if you hate it. I can change it easily to match the Atari game.

tomaswindsor:
It would be nice to be able to exclude some of the connected players from this feature. Ideally even from ever appearing in the group manager listing. (sometimes ppl use "virtual" player as a server and it's annoying to see arrow to this player all the time)
Also the arrows could be somewhat smaller (or have their size configurable).

PR #62, opened 2017-12-25
 Distant player's location is now hinted 

nschimme:
Implements support for  #17

This PR makes steps to support OpenGL 3.1 by replacing some deprecated API calls


nschimme:
I realized that GL_POLYGON has also been deprecated and is in the current code:
https://www.khronos.org/opengl/wiki/Legacy_OpenGL

I suppose this is outside of the scope of this PR

PR #88, opened 2018-01-15
Migrate away from QtIOCompressor for new saves

nschimme:
QtIOCompressor causes a lot of headaches and with Qt6 on the horizon I would rather not have it around if possible.

Let's migrate to the native `qCompress` and `qUncompress` methods which have better support.


nschimme:
@kalev your thoughts would be appreciated as well

kalev:
Sorry, I'm not familiar with the QtIOCompressor code.

nschimme:
no worries, guess I'll just be merging this then :-P :metal:

PR #91, opened 2018-01-18
Created a toggle to show hidden exit flags

nschimme:
Supported hidden exit flags:
* Deathtrap
* Underwater
* Oneway
* No Flee

Also, swim and direct sun exit flags are now shown in Exits/emulated


nschimme:
Naga had a suggestion to add `smob`

nschimme:
`smob` is actually quite spammy since its currently overloaded to include any major significant mob rather than loot dropping super mobs.

Issue #70, opened 2017-12-27
Introduce fall exit flag


nschimme:
Cyan → Exit leading into an air-room (where you instantly fall, such as at BBT entrance or Unqalome's Chasm Trail towards-center exits)

More details in https://github.com/MUME/MMapper/issues/40

Issue #69, opened 2017-12-27
Introduce blocked/guarded exit flag


nschimme:
Yellow → Blocked/guarded by a mobile (such as exit n from Gumak)

As mentioned in https://github.com/MUME/MMapper/issues/40

PR #95, opened 2018-01-19
Introduce `damage`, `fall`, and `guarded` exit flags

nschimme:
* These exit flags are colored as:
  * cyan --> `damage` and `fall`
  * yellow --> `guarded`

This implements issues #70 and #69


nschimme:
@tomaswindsor I've decided that `fall` and `damage` are close enough that they'll share the cyan color. Let me know if you disagree but folks will know the difference from the hidden exit flag hints.

Examples:
`Exits/emulated: north, {west}, up. - n:(fall)`
`Exits/emulated: north, south, west. - n:(damage) s:(damage).`

Issue #85, opened 2018-01-02
new line terminator parser/abstractparser.cpp


midoc:
The mmapper is sending \r\n regardless what the system will send for user typed input. So it is visible mmapper generated input.

nschimme:
We could add a AbstractParser boolean that detects a \r\n on user input and switches to display \r\n. This should also be reset if the player disconnects and reconnects to MMapper via the reset() method. This way we also do not need to make it visible by the user in the configuration.

nschimme:
@midoc it looks like MUME will always send \r\n no matter what the client types in. I tested both standard telnet (submits \r\n) and powwow (submits \n). Can you give me some specific examples?

midoc:
Yes it would be influenced also by application used. So in case i'm sitting on linux (and i would expect same can happen for mac as system where /r/n is not native) and i would type in powwow "open hiddenrock" i will send to mume line terminated with \n. In case i will type _open e. I will send "open hiddenrock" terminated with \r\n.  So you can distinguish between these two actions on mume server side :-).

btw there are also typos. So in code is \n\r, not \r\n.

nschimme:
@midoc I just want to verify that this issue is only for MMapper internal new line terminators to match the client newline terminator. You're OK with MUME returning `\r\n`s after room names or descriptions even if the client only terminates input with `\n`, correct?

midoc:
Yes, with obvious reason. Mmapper generated text towards MUME like _open xxx has its rule breaking scenario and user can be unaware that it can be tracked due this difference.

nschimme:
@midoc fixed the \n\r bug in https://github.com/MUME/MMapper/commit/2919a79d270f1bb078ba87f15f0b32e00aa2c7b6

PR #98, opened 2018-01-26
Reuse the same new line terminator when enhancing using input

nschimme:
Implements a fix for issue https://github.com/MUME/MMapper/issues/85


nschimme:
@midoc please let me know if you think this doesn't cover all bases

PR #109, opened 2018-05-05
WebSocket Secure support

nschimme:
In an effort to bring more security to MUME I'm adding WebSocket Secure support to MMapper. Other benefits include the ability to punch through firewalls which do packet inspection. This particular feature has been tested with the @waba4mume 's play-mume server: wss://test.waba.be:443/mume/play/websocket

I (possibly incorrectly) believe that `test.waba.be` a direct SSL connection to the `mume.org` server which means that this PR should introduce full encryption after a local client connects through MMapper over HTTPS to Websockify which then SSLs to MUME.

In order to use this feature, run this branch's MMapper and ensure your preferences have:
- Remote server: `test.waba.be`
- Remote port number: `443`
- [x] WebSocket Secure

I do not feel comfortable turning this feature on by default until we come up with the following solutions:
- Host Websockify directly on `mume.org`
- `mume.org` has HTTPS/WSS encryption through Let's Encrypt
   - For backwards compatibility we can use the sslh protocol demultiplexer: `sslh -p 443 --http localhost:80 --anyprot localhost:4242`
- Find a way to pass the client's IP through WebSockify to MUME such that their domain is exposed with `whois`


nschimme:
This PR is not required anymore with official TLS support now in MUME and MMapper. I'll leave the branch available if anyone needs it.

PR #174, opened 2019-12-28
Upgrade to Visual Studio 2019

nschimme:


nschimme:
There does not seem to be a point to doing this. The builds take longer.

PR #211, opened 2020-11-07
Changes to RoomUpdated flag

ethorondil:
This pull request has two commits, related to the room updated flag.

 (1) Prior to this, when the flag changes (if you walk into a room or update it manually), the display does not refresh.  This makes it refresh.
 (2) When you select multiple rooms, the checkbox is enabled so you can modify the room updated flag en masse.

The second feature is useful when TPing -- I can mark a whole region as non-updated, and see where I haven't been.

### DIFF
--- a/src/expandoracommon/room.cpp
+++ b/src/expandoracommon/room.cpp
@@ -17,7 +17,8 @@
 static constexpr const auto default_updateFlags = RoomUpdateFlags{}; /* none */
 static constexpr const auto mesh_updateFlags = RoomUpdateFlags{RoomUpdateEnum::Mesh};
 static constexpr const auto key_updateFlags = RoomUpdateFlags{RoomUpdateEnum::NodeLookupKey};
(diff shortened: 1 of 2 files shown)


nschimme:
Code style needs to be fixed:
https://travis-ci.com/github/MUME/MMapper/jobs/430483843

nschimme:
I'm a little worried about re-using this flag as a feature for TPing because it has an impact on updating rooms during mapping.

You can see the logic branches in room.cpp if you search for isUpToDate() which uses the `m_borked` flag. We called it borked because we weren't initially sure what it really represented and what the true impact of altering it is. It seems a little bit more clear now with the latest refactoring.

ethorondil:
Yeah, it's not the ideal way of doing TPing.  But it works pretty well, it's very easy to implement, and the feature required isn't actually specific to TPing -- to whatever extent you ever want to mark rooms as outdated/up to date, you might want to mark multiple rooms at once.

nschimme:
Thanks! Merged to master

PR #216, opened 2020-12-16
fix bug:  allow remote & local ports > 1000

cuantar:


cuantar:
... the commit message should say > 10k, obviously.  1k doesn't make sense in context.

cuantar:
will  fix commit message

PR #218, opened 2020-12-17
add _infomark command

cuantar:
add, update, or remove first text infomark without mouse

### DIFF
--- a/src/parser/AbstractParser-Commands.cpp
+++ b/src/parser/AbstractParser-Commands.cpp
@@ -21,6 +21,7 @@
 #include "../mapdata/DoorFlags.h"
 #include "../mapdata/ExitFlags.h"
 #include "../mapdata/enums.h"
+#include "../mapdata/infomark.h"
 #include "../mapdata/mmapper2room.h"
 #include "../syntax/SyntaxArgs.h"
 #include "../syntax/TreeParser.h"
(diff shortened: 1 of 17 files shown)


cuantar:
I'm looking at this again.  Currently there is an annoying bug that breaks this feature on my map (but not on the provided one):  Infomarks added by the _infomark command are only added to layer 0.  Instead, they should be added to whatever is the current layer.

cuantar:
The bug mentioned above is fixed, and infomarks now appear on the correct layer.

Still to do:  (1) add sub-commands like set/clear/type (e.g. herb, mob, comment, ...); and (2) migrate to new-style command parsing.

src/parser/AbstractParser-Commands.cpp
```diff
@@ -876,3 +888,70 @@ bool AbstractParser::evalSpecialCommandMap(StringView args)
     const auto matched = std::vector<StringView>{StringView{s}};
     return rec.callback(matched, args);
 }
+
+void AbstractParser::doCreateInfoMark(const StringView &cmd)
```
nschimme, line 892: I would prefer we use the new command syntax over the simple one. An example of this can be seen here: https://github.com/MUME/MMapper/commit/758461088fe885cfea5d85fcc87844957b18f45b

Issue #225, opened 2021-06-12
Room flag icons overlap; they should not.


cuantar:
One relatively low-tech way to solve this problem:  Edit the flag images so that the icons are not all in the same corner of the map tile; ensure that they still look good with multiple common flags (food, water, noride, eq, treasure, milk, etc.) enabled.  The icons for up/down directions may need their own locations where they will be undisturbed by flag images.

PR #228, opened 2021-06-13
modify flag icon art to be more useful

cuantar:
Most of the flag icons overlap in the upper left corner of the room tile. I've changed some of the load flags so that they are more understandable when multiple flags are active in the same room.

see issue #225


cuantar:
image

nschimme:
The up/down symbols seem like a nice improvement. I would prefer we allow any custom tile set as opposed to changing the look and feel of the basic textures. A simple solution would be following how Morrowind/Skyrim allow modding with a single directory that they look at for overriding default texture loading. Users could just drag/drop files into that directory with files that follow the expected naming convention that the textures are currently using within the .qrc file.

nschimme:
I have merged support for custom textures in https://github.com/MUME/MMapper/commit/3cecf26d8d891fb0ccabe1156546f791078ace40

Closing this since it can be handled that way. I'll see if I should take anything into master at a later date. I haven't had a chance to analyze the images too much yet.

Issue #232, opened 2021-08-19
Wildcard * search in Find Rooms


Mattias-Viklund:
Allowing for wildcard search in the Find Rooms dialog.

Example: "Majestic * tree"

PR #233, opened 2021-08-24
Wildcard search support in Find Rooms dialog.

Mattias-Viklund:


Mattias-Viklund:
#232

nschimme:
We migrated to C++17 so it's highly possible that GCC supports std::regex::multiline

This means that we could just add a boolean to allow pass through of the input regex. We wouldn't have to do any weird escaping then if this boolean is true.

nschimme:
Heh: `// REVISIT: Add another option to support regex where we do not santize the input`

Mattias-Viklund:
> 
> 
> Heh: `// REVISIT: Add another option to support regex where we do not santize the input`
I'll take a look at std::regex::multiline and also see about just making it a boolean.
C++ isn't my strongest suit though hah, but I'll see what I can do.

nschimme:
https://en.cppreference.com/w/cpp/regex/basic_regex/constants

The POSIX style regex is very different from what most people learn these days. If we switch to the multi-line we can simplify a lot.

Mattias-Viklund:
Yeah, that sounds like a very nice feature.
I'll hop on it.

nschimme:
https://github.com/MUME/MMapper/commit/2a86a688b81017f0eb61ffafac85bdd73b049484

PR #245, opened 2021-12-13
Add feature "Export MMapper2 XML Map As..." 

cosmos72:
As discussed on Discord chat,
MMapper2 currently lacks a mechanism to merge two maps that are derived from the same one,
and which differ only for some (relatively minor) modifications.

A typical use case is: a player has added some information (which he may not want to share,
or may not be relevant to other players, or simply have not been integrated yet into the upstream map)
to an upstream map, but in the meantime a new upstream map was published,
and he wants to merge his personal changes into a copy of the new upstream map.

This is a classical 3-way merge, with a base map and two other maps derived from it.

MMapper2 currently has no support for such a use case, and it may contain conflicts between the two derived maps,
which may be difficult or impossible to resolve without user input.

**This pull request adds the following feature: allow exporting a map to human-readable, text (actually XML) format,**
named ".mm2xml", which empowers users to use text editing tools on MMapper2 maps, such as `diff(1)`, `patch(1)` or even plain text editors.

It also allows maintaining patches on top of an existing map, for example using `diff(1)`

A subsequent pull request will add the ability to **import** (i.e. load) maps saved in the ".mm2xml" format.

### DIFF
--- a/src/mainwindow/mainwindow.cpp
+++ b/src/mainwindow/mainwindow.cpp
@@ -545,11 +545,11 @@ void MainWindow::createActions()
     saveAsAct->setStatusTip(tr("Save the document under a new name"));
     connect(saveAsAct, &QAction::triggered, this, &MainWindow::slot_saveAs);
 
-    exportBaseMapAct = new QAction(tr("Export &Base Map As..."), this);
+    exportBaseMapAct = new QAction(tr("Export MMapper2 &Base Map As..."), this);
     exportBaseMapAct->setStatusTip(tr("Save a copy of the map with no secrets"));
     connect(exportBaseMapAct, &QAction::triggered, this, &MainWindow::slot_exportBaseMap);
 
-    exportMm2xmlMapAct = new QAction(tr("Export MM2&XML Map As..."), this);
+    exportMm2xmlMapAct = new QAction(tr("Export MMapper2 &XML Map As..."), this);
     exportMm2xmlMapAct->setStatusTip(tr("Save a copy of the map in the MM2XML format"));
     connect(exportMm2xmlMapAct, &QAction::triggered, this, &MainWindow::slot_exportMm2xmlMap);
 
--- a/src/mainwindow/mainwindow.cpp
+++ b/src/mainwindow/mainwindow.cpp
@@ -50,6 +50,7 @@
 #include "../mapfrontend/mapfrontend.h"
 #include "../mapstorage/MmpMapStorage.h"
 #include "../mapstorage/PandoraMapStorage.h"
+#include "../mapstorage/XmlMapStorage.h"
 #include "../mapstorage/abstractmapstorage.h"
 #include "../mapstorage/filesaver.h"
 #include "../mapstorage/jsonmapstorage.h"
@@ -547,6 +548,10 @@ void MainWindow::createActions()
     exportBaseMapAct = new QAction(tr("Export &Base Map As..."), this);
     exportBaseMapAct->setStatusTip(tr("Save a copy of the map with no secrets"));
     connect(exportBaseMapAct, &QAction::triggered, this, &MainWindow::slot_exportBaseMap);
+
+    exportMm2xmlMapAct = new QAction(tr("Export MM2&XML Map As..."), this);
+    exportMm2xmlMapAct->setStatusTip(tr("Save a copy of the map in the MM2XML format"));
+    connect(exportMm2xmlMapAct, &QAction::triggered, this, &MainWindow::slot_exportMm2xmlMap);
 
     exportWebMapAct = new QAction(tr("Export &Web Map As..."), this);
     exportWebMapAct->setStatusTip(tr("Save a copy of the map for webclients"));
@@ -1062,6 +1067,7 @@ void MainWindow::disableActions(bool value)
     reloadAct->setDisabled(value);
     saveAsAct->setDisabled(value);
     exportBaseMapAct->setDisabled(value);
+    exportMm2xmlMapAct->setDisabled(value);
     exportWebMapAct->setDisabled(value);
     exportMmpMapAct->setDisabled(value);
     exitAct->setDisabled(value);
@@ -1111,6 +1117,7 @@ void MainWindow::setupMenuBar()
     fileMenu->addSeparator();
     QMenu *exportMenu = fileMenu->addMenu(QIcon::fromTheme("document-send"), tr("&Export"));
     exportMenu->addAction(exportBaseMapAct);
+    exportMenu->addAction(exportMm2xmlMapAct);
     exportMenu->addAction(exportWebMapAct);
     exportMenu->addAction(exportMmpMapAct);
     fileMenu->addAction(mergeAct);
@@ -1662,6 +1669,33 @@ bool MainWindow::slot_exportBaseMap()
     return saveFile(fileNames[0], SaveModeEnum::BASEMAP, SaveFormatEnum::MM2);
 }
 
+bool MainWindow::slot_exportMm2xmlMap()
+{
+    const auto makeSaveDialog = [this]() {
+        // FIXME: code duplication
+        auto save = std::make_unique<QFileDialog>(this,
+                                                  "Choose map file name ...",
+                                                  QDir::current().absolutePath());
+        save->setFileMode(QFileDialog::AnyFile);
+        save->setDirectory(QDir(getConfig().autoLoad.lastMapDirectory));
+        save->setNameFilter("MM2XML Maps (*.mm2xml)");
+        save->setDefaultSuffix("mm2xml");
+        save->setAcceptMode(QFileDialog::AcceptSave);
+        save->selectFile(QFileInfo(m_mapData->getFileName())
+                             .fileName()
+                             .replace(QRegularExpression(R"(\.mm2$)"), ".mm2xml"));
+
+        return save;
+    };
+
+    const auto fileNames = getSaveFileNames(makeSaveDialog());
+    if (fileNames.isEmpty()) {
+        statusBar()->showMessage(tr("No filename provided"), 2000);
+        return false;
+    }
+    return saveFile(fileNames[0], SaveModeEnum::FULL, SaveFormatEnum::MM2XML);
+}
+
 bool MainWindow::slot_exportWebMap()
 {
     const auto makeSaveDialog = [this]() {
@@ -1861,6 +1895,8 @@ bool MainWindow::saveFile(const QString &fileName,
         switch (format) {
         case SaveFormatEnum::MM2:
             return std::make_unique<MapStorage>(*m_mapData, fileName, &saver.file(), this);
+        case SaveFormatEnum::MM2XML:
+            return std::make_unique<XmlMapStorage>(*m_mapData, fileName, &saver.file(), this);
         case SaveFormatEnum::MMP:
             return std::make_unique<MmpMapStorage>(*m_mapData, fileName, &saver.file(), this);
         case SaveFormatEnum::WEB:
--- a/src/mapstorage/XmlMapStorage.cpp
+++ b/src/mapstorage/XmlMapStorage.cpp
@@ -242,45 +242,6 @@ void XmlMapStorage::log(const QString &msg)
     emit sig_log("XmlMapStorage", msg);
 }
 
-#define DECL(X, ...) #X,
-static const char *const alignNames[] = {
-    X_FOREACH_RoomAlignEnum(DECL) //
-};
-static const char *const doorFlagNames[] = {
-    X_FOREACH_DOOR_FLAG(DECL) //
-};
-static const char *const exitFlagNames[] = {
-    X_FOREACH_EXIT_FLAG(DECL) //
-};
-static const char *const lightNames[] = {
-    X_FOREACH_RoomLightEnum(DECL) //
-};
-static const char *const loadFlagNames[] = {
-    X_FOREACH_ROOM_LOAD_FLAG(DECL) //
-};
-static const char *const markClassNames[] = {
-    X_FOREACH_INFOMARK_CLASS(DECL) //
-};
-static const char *const markTypeNames[] = {
-    X_FOREACH_INFOMARK_TYPE(DECL) //
-};
-static const char *const mobFlagNames[] = {
-    X_FOREACH_ROOM_MOB_FLAG(DECL) //
-};
-static const char *const portableNames[] = {
-    X_FOREACH_RoomPortableEnum(DECL) //
-};
-static const char *const ridableNames[] = {
-    X_FOREACH_RoomRidableEnum(DECL) //
-};
-static const char *const sundeathNames[] = {
-    X_FOREACH_RoomSundeathEnum(DECL) //
-};
-static const char *const terrainNames[] = {
-    X_FOREACH_RoomTerrainEnum(DECL) //
-};
-#undef DECL
-
 void XmlMapStorage::saveDoorFlags(QXmlStreamWriter &stream, DoorFlags fl)
 {
     if (fl.isEmpty()) {
@@ -288,11 +249,11 @@ void XmlMapStorage::saveDoorFlags(QXmlStreamWriter &stream, DoorFlags fl)
     }
     QString list;
     const char *separator = "";
-    for (uint x = 0; x < sizeof(doorFlagNames) / sizeof(doorFlagNames[0]); x++) {
+    for (uint x = 0; x < NUM_DOOR_FLAGS; x++) {
         if (fl.contains(DoorFlagEnum(x))) {
             list += separator;
             separator = ",";
-            list += doorFlagNames[x];
+            list += doorFlagName(DoorFlagEnum(x));
         }
     }
     saveXmlAttribute(stream, "doorflags", list);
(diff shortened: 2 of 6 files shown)


src/mapstorage/XmlMapStorage.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+NODISCARD QString XmlMapStorage::lightName(const RoomLightEnum x)
+{
+    return enumName(uint(x), lightNames, sizeof(lightNames) / sizeof(lightNames[0]));
+}
+
+NODISCARD QString XmlMapStorage::markClassName(const InfoMarkClassEnum e)
+{
+    const uint x = uint(e);
+    const uint count = sizeof(markClassNames) / sizeof(markClassNames[0]);
```
nschimme, line 370: Why would there be more mark classes than what the code is aware of? This could mean map corruption!

src/mapstorage/XmlMapStorage.cpp
```diff
@@ -0,0 +1,420 @@
...
+            }
+            stream.writeCharacters(mobFlagNames[x]);
+        }
+    }
+    stream.writeEndElement();
+}
+
+NODISCARD QString XmlMapStorage::alignName(const RoomAlignEnum x)
+{
+    return enumName(uint(x), alignNames, sizeof(alignNames) / sizeof(alignNames[0]));
```
nschimme, line 359: Why even generate alignNames? We could just have that X_FOREACH macro here and return the value

src/mapstorage/XmlMapStorage.cpp
```diff
@@ -0,0 +1,420 @@
...
+    }
+    stream.writeStartElement("flags");
+    QString separator;
+    for (uint x = 0; x < sizeof(loadFlagNames) / sizeof(loadFlagNames[0]); x++) {
+        if (fl.contains(RoomLoadFlagEnum(x))) {
+            stream.writeCharacters(separator);
+            if (separator.isEmpty()) {
+                separator = ",";
+            }
+            stream.writeCharacters(loadFlagNames[x]);
```
nschimme, line 332: Why not have each flag be a separate XML element?

src/mapstorage/XmlMapStorage.cpp
cosmos72: My habit at defensive programming.

Would you prefer an assertion, a throw, or something else?

src/mapstorage/XmlMapStorage.cpp
```diff
@@ -0,0 +1,420 @@
...
+    progressCounter.reset();
+    progressCounter.increaseTotalStepsBy(saver.getRoomsCount()
+                                         + static_cast<quint32>(markerList.size()));
+
+    QXmlStreamWriter stream(m_file);
+    stream.setAutoFormatting(true);
+    stream.writeStartDocument();
+
+    stream.writeStartElement("map");
+    stream.writeAttribute("type", "mmapper2xml");
```
nschimme, line 81: Is there any way we could provide an official XML DTD as well? This would help validate things.

src/mapstorage/XmlMapStorage.cpp
cosmos72: Well, I could use a `switch (x)` and a `X_FOREACH_*` macro inside it expanding to each case.
I guess a decent compiler would produce basically the same code.

src/mapstorage/XmlMapStorage.cpp
cosmos72: That's certainly possible. It could possibly generate lots of lines though.
But on the other hand it would be much easier to validate with a DTD or XML Schema (see next item)

src/mapstorage/XmlMapStorage.cpp
cosmos72: Sure, I can prepare that :)
I prefer XML schemas, but a DTD wouldn't be a problem too - we could have both

src/mapstorage/XmlMapStorage.cpp
nschimme: assert() would be an acceptable solution. We shouldn't persist the unknown value to the XML file though. This is where a better schema validator (like a DTD) will help us.

I think we would need to be more paranoid during data loading. We would most likely want to abort at that point and inform the user they corrupted the data.

src/mapstorage/XmlMapStorage.cpp
nschimme: Yeah, plus the macro since X_FOREACH is used to even generate the enum itself so it's extremely safe.

src/mapstorage/XmlMapStorage.cpp
cosmos72: Done. RoomLoadFlags, DoorFlags and ExitFlags are now saved as repeated XML elements instead of a comma-separated list inside the value of an XML attribute

src/mapstorage/XmlMapStorage.cpp
cosmos72: Done.

src/mapstorage/XmlMapStorage.cpp
cosmos72: I added `assert(false)` at the end of every `switch (e)` to catch "impossible" enum values.

src/mainwindow/mainwindow.cpp
```diff
@@ -1662,6 +1669,33 @@ bool MainWindow::slot_exportBaseMap()
...
+                                                  "Choose map file name ...",
+                                                  QDir::current().absolutePath());
+        save->setFileMode(QFileDialog::AnyFile);
+        save->setDirectory(QDir(getConfig().autoLoad.lastMapDirectory));
+        save->setNameFilter("MM2XML Maps (*.mm2xml)");
+        save->setDefaultSuffix("mm2xml");
+        save->setAcceptMode(QFileDialog::AcceptSave);
+        save->selectFile(QFileInfo(m_mapData->getFileName())
+                             .fileName()
+                             .replace(QRegularExpression(R"(\.mm2$)"), ".mm2xml"));
```
nschimme, line 1686: This might not work in the future when we load a .mm2xml file and then try to export that same file again. We might need to fix this bug (and code duplication) in other parts of the code too.

src/mapstorage/XmlMapStorage.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+    // Collect the room and marker lists. The room list can't be acquired
+    // directly apparently and we have to go through a RoomSaver which receives
+    // them from a sort of callback function.
+    // The RoomSaver acts as a lock on the rooms.
+    ConstRoomList roomList;
+    roomList.reserve(m_mapData.getRoomsCount());
+
+    RoomSaver saver(m_mapData, roomList);
+    for (uint i = 0, n = m_mapData.getRoomsCount(); i < n; ++i) {
```
nschimme, line 66: Why not just cache `n` as `roomsCount` higher up in the code? You're already calling it several times and there is no need to "optimize" here in the loop really.

src/mainwindow/mainwindow.cpp
cosmos72: In my idea, `m_mapData->getFileName()` would always be a `.mm2` file, never a `.mm2xml` file.
From another perspective, loading `.mm2xml` file in my opinion should **not** set `m_mapData->getFileName()`

Does it make sense?

src/mapstorage/XmlMapStorage.cpp
cosmos72: makes sense

src/mapstorage/XmlMapStorage.cpp
```diff
@@ -0,0 +1,462 @@
...
+    stream.writeAttribute("z", QString("%1").arg(pos.z));
+    stream.writeEndElement(); // end coord
+}
+
+void XmlMapStorage::saveExit(QXmlStreamWriter &stream, const Exit &e, ExitDirEnum dir)
+{
+    if (!e.exitIsExit() || e.outIsEmpty()) {
+        return;
+    }
+    stream.writeStartElement("exit");
```
nschimme, line 162: Should there be a parent element called "exits" for all of the exits?

src/mapstorage/XmlMapStorage.cpp
```diff
@@ -0,0 +1,462 @@
...
+    const char *separator = nullptr;
+    for (const RoomId id : e.outRange()) {
+        if (separator == nullptr) {
+            separator = ",";
+        } else {
+            idlist += separator;
+        }
+        idlist += QString("%1").arg(id.asUint32());
+    }
+    stream.writeAttribute("to", idlist);
```
nschimme, line 183: Just make "to" an element?

src/mapstorage/XmlMapStorage.cpp
```diff
@@ -0,0 +1,462 @@
...
+    }
+}
+
+void XmlMapStorage::saveExitFlags(QXmlStreamWriter &stream, ExitFlags fl)
+{
+    fl.remove(ExitFlagEnum::EXIT); // always set, do not save it
+    if (fl.isEmpty()) {
+        return;
+    }
+    for (uint x = 0; x < NUM_EXIT_FLAGS; x++) {
```
nschimme, line 262: You can use X_FOREACH_EXIT_FLAG instead

src/mapstorage/XmlMapStorage.cpp
cosmos72: And repeat &lt;to&gt;roomId&lt;/to&gt; for each connected room?
That's surely possible, but again quite redundant.
Also, one exit connecting to multiple rooms is rare, basically only happens in random mazes.

It could be useful if the DTD or XML Schema allows XML validator software to cross-reference exits with room ids, i.e. to automatically check if the target room id really exists in the map.
But the only DTD type that allows such a feature is the ID type, which cannot be numeric.

All in all, I'm not sure the extra verbosity is worth it in this case... but if you want I'll make the change

src/mapstorage/XmlMapStorage.cpp
```diff
@@ -0,0 +1,462 @@
...
+{
+    emit sig_log("XmlMapStorage", msg);
+}
+
+void XmlMapStorage::saveDoorFlags(QXmlStreamWriter &stream, DoorFlags fl)
+{
+    if (fl.isEmpty()) {
+        return;
+    }
+    for (uint x = 0; x < NUM_DOOR_FLAGS; x++) {
```
nschimme, line 248: The X_FOREACH_DOOR_FLAG macro would work here too

src/mapstorage/XmlMapStorage.cpp
nschimme: Another alternative are the enums in https://github.com/MUME/MMapper/blob/master/src/mapdata/enums.h

cosmos72: I am not sure to understand how you propose to write the loop on all `ExitFlagEnum` values.
Maybe you mean something like
```C++
#define X(UPPER, ...) ExitFlagEnum::UPPER,
for (ExitFlagEnum e : { X_FOREACH_EXIT_FLAG(x) }) {
    // ...
}
#undef X
```
using an implicit `std::initializer_list<ExitFlagEnum>` ?
Possible, but the heavy use of macros in C++ is really against my usual style...

src/mapstorage/XmlMapStorage.cpp
```diff
@@ -0,0 +1,462 @@
...
+#define CASE(X) \
+    case RoomPortableEnum::X: \
+        return #X;
+        switch (e) {
+            X_FOREACH_RoomPortableEnum(CASE)
+        }
+#undef CASE
+        assert(false); // reachable only on invalid RoomPortableEnum
+    }
+    return "";
```
nschimme, line 416: Doesn't this mean we're going to persist an empty string to the file if it's undefined? Shouldn't we just omit the element entirely?

cosmos72: We omit the element entirely.
The empty string is detected in `XmlMapStorage::saveXmlElement` and `XmlMapStorage::saveXmlAttribute`
which omit the entire element (or attribute) if the value to save is the empty string.

src/mapstorage/XmlMapStorage.cpp
cosmos72: I'll check how to use the macros `ALL_EXIT_FLAGS` and friends, as per your suggestion. If they allow writing
```
for (ExitFlagEnum e : ALL_EXIT_FLAGS) {
    // ...
}
```
then it's definitely cleaner :)

src/mapstorage/XmlMapStorage.cpp
cosmos72: same considerations as previous point above

src/mapstorage/XmlMapStorage.cpp
nschimme: Hrm, I do agree with the <rooms> idea. What if intead of <to> we have <exits> has a bunch of <exit> that can be redundant? There is some duplication but its still better than a ton of <to> elements.

For example:
```
<exits>
<exit dir="n" doorname="door" to="1">
<exit dir="n" doorname="door" to="2">
<exit dir="e" to="3">
</exits>
```

src/mapstorage/XmlMapStorage.cpp
nschimme: Yeah, those kind of things probably don't belong in a DTD anyway. We can do that during map load time.

src/mapstorage/XmlMapStorage.cpp
nschimme: See if the enums work better. They haven't been used very much in the codebase yet but they exist for this purpose :-)

src/mapstorage/XmlMapStorage.cpp
nschimme: How about using std::optional instead of an empty string? It's more clear then how it will be used

src/mainwindow/mainwindow.cpp
cosmos72: Let me rephrase my idea: when loading a `.mm2xml` map, the map name or filename is **not** set in MMapper2.

After all, it's an import: users are supposed to save it back in `.mm2` format, and **that** will set the map name and filename.

src/mapstorage/XmlMapStorage.cpp
cosmos72: If you want, ok.
I thought the empty string was already enough a special value.

src/mapstorage/XmlMapStorage.cpp
nschimme: Sure, I was just trying to suggest something that would optimize on disk space given you had expressed some concerns. I'm fine with the verbosity since it will also make diffs easier to interpret.

src/mapstorage/XmlMapStorage.cpp
cosmos72: done

src/mapstorage/XmlMapStorage.cpp
cosmos72: As agreed (hoping I correctly understood your preference), I made "to" an element, i.e.
```xml
<exit dir="north">
  <to>3</to>
  <to>4</to>
</exit>
```
so now there's no need to store a comma-separated list of room ids in a single "to" attribute

src/mapstorage/XmlMapStorage.cpp
cosmos72: Same as above, "to" is now a repeatable element.

src/mapstorage/XmlMapStorage.cpp
cosmos72: Done for `DoorFlagEnum`, `ExitFlagEnum`, `RoomLoadFlagEnum` and `RoomMobFlagEnum`.
The new iteration
```C++
for (const ExitFlagEnum e : ALL_EXIT_FLAGS) {
 // ...
}
```
works and is indeed cleaner

src/mapstorage/XmlMapStorage.cpp
cosmos72: done too, see previous point above

src/mapstorage/XmlMapStorage.cpp
cosmos72: Looking better, the lines
```C++
    saveXmlElement(stream, "description", room.getDescription().toQString());
    saveXmlElement(stream, "contents", room.getContents().toQString());
    saveXmlElement(stream, "note", room.getNote().toQString());
```
would become more complicated, because for those we want to treat these empty string as "do not save the element", which the current `XmlMapStorage::saveXmlElement` does automatically.

If `XmlMapStorage::saveXmlElement` and `XmlMapStorage::saveXmlAttribute` are changed to accept `std::optional<QString>` with the semantics that only `std::nullopt` omits the element entirely,
then the code above must check for empty TaggedString, i.e.

```C++
    if (!room.getDescription().isEmpty()) {
        saveXmlElement(stream, "description", room.getDescription().toQString());
    }
    if (!room.getContents().isEmpty()) {
        saveXmlElement(stream, "contents", room.getContents().toQString());
    }
    if (!room.getNote().isEmpty()) {
        saveXmlElement(stream, "note", room.getNote().toQString());
    }
```
which defeats the very purpose of having `saveXmlElement()` accept a `std::optional<QString>`

In other words, the `.mm2xml` format already treats all empty strings specially - by **not** storing the element or attribute containing only them (or **should** treat empty strings in this way: I may have missed some cases, which in my opinition should be fixed)
so I do not see much need to introduce a different special value - an empty `std::optional<QString>` - to indicate the same concept.

src/mapstorage/XmlMapStorage.cpp
nschimme: Looks good. Very few rooms have multiple exits so this might be a little verbose but it's not the end of the world.

src/mapstorage/XmlMapStorage.cpp
nschimme: Agreed, seems like the best solution then. Thanks for digging into it 👍

nschimme:
Looks good to me. Do you want to squash this into a single commit and then start working on the loading?

cosmos72:
sure :) proceeding.

PR #248, opened 2021-12-28
Add "Room Panel" feature

### LINKED ISSUE #238: Proposal: new panel keeping track of mobs and players present in the current room
I am proposing a new feature, and if accepted I am also offering to implement it myself:

Add a new Mob/Player Panel, somewhat similar to the already existing Group Panel, that shows the list of mobs and players present in the current room.
The idea is to keep it up-to-date to the best of MMapper abilities, i.e. updating it when:
- the player moves
- some mob or player enter or leaves the room, provided that the player can see it (i.e. if MUME sends the message "Foo leaves north/south/east/west/up/down")

And also taking into account that movement of other mobs/players in the same [group](https://mume.org/help/group) as the player are omitted by MUME in many cases - for example the leader of a group receives a message like
```ACK! A pack horse didn't follow you, you lost him.```
when he/she loses a follower, but in nospam mode MUME will **not** tell if a follower successfully follows his/her leader.

Another useful feature could be:
- showing in the mob/player list what **each** one is doing, who he/she is attacking etc.

i.e. the panel is basically intended to be a snapshot of `look` or `examine`, and also updated in real-time to the best of MMapper abilities when the player or somebody else moves, or someone in the room does something significant (attacking, casting, resting, standing etc).

I am interested in feedback for the proposal above, in particular:
- does it fit MMapper features and functions, or it would be better implemented in some other program?
- does it sound useful to MUME players, or it would need some minor/major changes to be useful?

My hope is that such feature could be **useful in situations with a lot of mobs/players** - **including PK situations** - where **messages scroll very quickly and it's difficult to keep track** of who is in the room and what they are doing.

### PR DESCRIPTION
**Pull request by cosmos72**
Fixes #238 - implement "Room Panel" showing GMCP messages Room.Chars.* i.e. show in a new dedicated panel
the list of mobs/players in current room and who they are fighting

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -354,6 +354,14 @@ set(mmapper_SRCS
     proxy/proxy.h
     proxy/telnetfilter.cpp
     proxy/telnetfilter.h
+    roompanel/RoomManager.cpp
+    roompanel/RoomManager.h
+    roompanel/RoomMob.cpp
+    roompanel/RoomMob.h
+    roompanel/RoomMobs.cpp
+    roompanel/RoomMobs.h
+    roompanel/RoomWidget.cpp
+    roompanel/RoomWidget.h
     syntax/Accept.cpp
     syntax/Accept.h
     syntax/IArgument.cpp
--- a/src/pandoragroup/CGroupChar.cpp
+++ b/src/pandoragroup/CGroupChar.cpp
@@ -8,6 +8,7 @@
 #include <QDebug>
 #include <QMessageLogContext>
 
+#include "../global/QuotedQString.h"
 #include "../global/roomid.h"
 #include "../parser/abstractparser.h"
 
@@ -49,25 +50,6 @@ const QVariantMap CGroupChar::toVariantMap() const
     root[playerDataKey] = playerData;
     return root;
 }
-
-// TODO: move this somewhere more appropriate
-struct NODISCARD QuotedQString final
-{
-private:
-    QString str;
-
-public:
-    explicit QuotedQString(QString input)
-        : str{std::move(input)}
-    {}
-    friend QDebug &operator<<(QDebug &debug, const QuotedQString &q)
-    {
-        debug.quote();
-        debug << q.str;
-        debug.noquote();
-        return debug;
-    }
-};
 
 bool CGroupChar::updateFromVariantMap(const QVariantMap &data)
 {
--- a/src/proxy/GmcpMessage.h
+++ b/src/proxy/GmcpMessage.h
@@ -31,6 +31,10 @@ class ParseEvent;
       MmapperCommGroupTell, \
       "mmapper.comm.grouptell", \
       "MMapper.Comm.GroupTell") \
+    X(ROOM_CHARS_ADD, RoomCharsAdd, "room.chars.add", "Room.Chars.Add") \
+    X(ROOM_CHARS_REMOVE, RoomCharsRemove, "room.chars.remove", "Room.Chars.Remove") \
+    X(ROOM_CHARS_SET, RoomCharsSet, "room.chars.set", "Room.Chars.Set") \
+    X(ROOM_CHARS_UPDATE, RoomCharsUpdate, "room.chars.update", "Room.Chars.Update") \
     /* define gmcp message types above */
 
 enum class NODISCARD GmcpMessageTypeEnum {
@@ -40,9 +44,8 @@ enum class NODISCARD GmcpMessageTypeEnum {
 #undef X_DECL_GMCP_MESSAGE_TYPE
 };
 
-static constexpr const size_t NUM_GMCP_MESSAGES = 10u;
-static_assert(NUM_GMCP_MESSAGES
-              == static_cast<int>(GmcpMessageTypeEnum::MMAPPER_COMM_GROUPTELL) + 1);
+static constexpr const size_t NUM_GMCP_MESSAGES = 14u;
+static_assert(NUM_GMCP_MESSAGES == static_cast<int>(GmcpMessageTypeEnum::ROOM_CHARS_UPDATE) + 1);
 DEFINE_ENUM_COUNT(GmcpMessageTypeEnum, NUM_GMCP_MESSAGES)
 
 namespace tags {
--- a/src/roompanel/RoomManager.cpp
+++ b/src/roompanel/RoomManager.cpp
@@ -1,0 +1,235 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (C) 2021 The MMapper Authors
+// Author: Massimiliano Ghilardi <massimiliano.ghilardi@gmail.com> (Cosmos)
+
+#include <memory>
+#include <string>
+#include <utility>
+#include <QtCore>
+
+#include "../global/StringView.h"
+#include "../proxy/GmcpMessage.h"
+#include "RoomManager.h"
+#include "RoomMobs.h"
+
+RoomManager::RoomManager(QObject *const parent)
+    : QObject{parent}
+    , m_room{this}
+    , m_debug{false}
+{}
+
+RoomManager::~RoomManager() = default;
+
+void RoomManager::slot_reset()
+{
+    m_room.resetMobs();
+}
+
+void RoomManager::updateWidget()
+{
+    emit sig_updateWidget();
+}
+
+void RoomManager::slot_parseGmcpInput(const GmcpMessage &msg)
+{
+    if (msg.isRoomCharsAdd())
+        parseGmcpAdd(msg);
+    else if (msg.isRoomCharsRemove())
+        parseGmcpRemove(msg);
+    else if (msg.isRoomCharsSet())
+        parseGmcpSet(msg);
+    else if (msg.isRoomCharsUpdate())
+        parseGmcpUpdate(msg);
+}
+
+void RoomManager::parseGmcpAdd(const GmcpMessage &msg)
+{
+    showGmcp(msg);
+    QJsonDocument doc = QJsonDocument::fromJson(msg.getJson()->toQString().toUtf8());
+    if (!doc.isObject()) {
+        if (m_debug) {
+            qWarning().noquote() << "RoomManager received GMCP" << msg.getName().toQString()
+                                 << "containing invalid Json: expecting object, got"
+                                 << (msg.getJson() ? msg.getJson()->toQString() : QString());
+        }
+        return;
+    }
+    addMob(doc.object());
+}
+
+void RoomManager::parseGmcpRemove(const GmcpMessage &msg)
+{
+    showGmcp(msg);
+    // payload is a single number (usually followed by a space), not an array or object
+    // thus parsed msg.getDocument() is not valid: use msg.getJson() instead
+    if (!msg.getJson() || msg.getJson()->getStdString().empty()) {
+        qWarning().noquote() << "RoomManager received GMCP" << msg.getName().toQString()
+                             << "containing invalid empty payload: expecting number";
+        return;
+    }
+    const std::string &str = msg.getJson()->getStdString();
+    size_t pos = 0;
+    unsigned long num = 0;
+    bool bad = false;
+    try {
+        num = std::stoul(str, &pos, 10);
+    } catch (...) {
+        bad = true; // str does not contain a valid unsigned long
+    }
+    const RoomMob::Id id = static_cast<RoomMob::Id>(num);
+    if (bad || id == RoomMobData::NOID || static_cast<unsigned long>(id) != num
+        || (pos < str.size() && str[pos] > ' ')) {
+        qWarning().noquote() << "RoomManager received GMCP" << msg.getName().toQString()
+                             << "containing invalid payload: expecting unsigned number, got"
+                             << msg.getJson()->toQString();
+        return;
+    }
+    if (m_room.removeMobById(id)) {
+        updateWidget();
+    }
+}
+
+void RoomManager::parseGmcpSet(const GmcpMessage &msg)
+{
+    showGmcp(msg);
+    QJsonDocument doc = QJsonDocument::fromJson(msg.getJson()->toQString().toUtf8());
+    if (!doc.isArray()) {
+        if (m_debug) {
+            qWarning().noquote() << "RoomManager received GMCP" << msg.getName().toQString()
+                                 << "containing invalid Json: expecting array, got"
+                                 << (msg.getJson() ? msg.getJson()->toQString() : QString());
+        }
+        return;
+    }
+    m_room.resetMobs();
+    for (QJsonValueRef value : doc.array()) {
+        if (value.isObject()) {
+            addMob(value.toObject());
+        } else {
+            if (m_debug) {
+                qWarning().noquote()
+                    << "RoomManager received GMCP" << msg.getName().toQString()
+                    << "containing invalid Json: expecting array of objects, got [/*...*/"
+                    << value.type() << "/*...*/]";
+            }
+        }
+    }
+    updateWidget();
+}
+
+void RoomManager::parseGmcpUpdate(const GmcpMessage &msg)
+{
+    showGmcp(msg);
+    QJsonDocument doc = QJsonDocument::fromJson(msg.getJson()->toQString().toUtf8());
+    if (!doc.isObject()) {
+        if (m_debug) {
+            qWarning().noquote() << "RoomManager received GMCP" << msg.getName().toQString()
+                                 << "containing invalid Json: expecting object, got"
+                                 << (msg.getJson() ? msg.getJson()->toQString() : QString());
+        }
+        return;
+    }
+    updateMob(doc.object());
+}
+
+void RoomManager::addMob(const QJsonObject &obj)
+{
+    RoomMobUpdate data;
+    if (toMob(obj, data)) {
+        m_room.addMob(std::move(data));
+        updateWidget();
+    }
+}
+
+void RoomManager::updateMob(const QJsonObject &obj)
+{
+    RoomMobUpdate data;
+    if (toMob(obj, data)) {
+        if (m_room.updateMob(std::move(data))) {
+            updateWidget();
+        }
+    }
+}
+
+inline void RoomManager::showGmcp(const GmcpMessage &msg) const
+{
+    if (m_debug) {
+        qDebug().noquote() << "RoomManager received GMCP:" << msg.toRawBytes();
+    }
+}
+
+const QHash<QString, MobFieldEnum> RoomManager::mobFields{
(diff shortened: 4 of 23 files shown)

### CODE REVIEW

File: src/CMakeLists.txt [hunk 1]
```diff
@@ -352,6 +352,14 @@ set(mmapper_SRCS
     proxy/proxy.h
     proxy/telnetfilter.cpp
     proxy/telnetfilter.h
+    roompanel/CRoom.cpp
```
nschimme, line 355: The prefix of "C" is a PandoraMapper naming anti-pattern. We're starting to converge on CamelCase naming convention for files

src/roompanel/roomwidget.cpp
```diff
@@ -0,0 +1,252 @@
...
+        const int row = index.row();
+        const int column = index.column();
+
+        switch (role) {
+        case Qt::DisplayRole:
+            return getMobField(row, column);
+        case Qt::BackgroundRole: {
+            const QColor color = getMobColor(row);
+            if (color.isValid()) {
+                // honor CRoomMob color - hilights enemies
```
nschimme, line 91: highlight :-)

src/CMakeLists.txt
cosmos72: fixed

src/roompanel/roomwidget.cpp
cosmos72: fixed too :-)

src/pandoragroup/mmapper2group.cpp
```diff
@@ -564,7 +564,7 @@ void Mmapper2Group::slot_parseGmcpInput(const GmcpMessage &msg)
 
     if (msg.isCharVitals()) {
         // "Char.Vitals {\"hp\":100,\"maxhp\":100,\"mana\":100,\"maxmana\":100,\"mp\":139,\"maxmp\":139}"
-        QJsonDocument doc = QJsonDocument::fromJson(msg.getJson()->toQString().toUtf8());
+        const QJsonDocument &doc = msg.getDocument();
```
nschimme, line 567: Is there a reason we decided to pre-parse the JSON? The main reason I did it in a lazy fashion was because MMapper doesn't care about all GMCP messages and we could avoid some deserialization overhead.

Also, if MMapper only deserializes the message once then we aren't really gaining anything by doing it ahead of time. I assume this is the case here.

src/display/mapcanvas.cpp
```diff
@@ -1077,7 +1077,7 @@ void MapCanvas::userPressedEscape(bool /*pressed*/)
 
     case CanvasMouseModeEnum::MOVE:
         // special case for move: right click selects infomarks
-        FALLTHRU;
+        FALLTHROUGH;
```
nschimme, line 1080: We should extract this out and package it as a separate PR enhancement

src/global/Flags.h
```diff
@@ -179,23 +179,23 @@ class NODISCARD Flags
     }
 
 public:
-    NODISCARD inline bool contains(const Flag flag) const noexcept
+    NODISCARD inline constexpr bool contains(const Flag flag) const noexcept
```
nschimme, line 182: These constexpr enhancements should also be a separate PR

src/pandoragroup/CGroupChar.cpp
```diff
@@ -50,25 +51,6 @@ const QVariantMap CGroupChar::toVariantMap() const
     return root;
 }
 
-// TODO: move this somewhere more appropriate
```
nschimme, line 53: Thanks!

src/proxy/GmcpMessage.h
```diff
@@ -62,9 +66,14 @@ using GmcpJson = TaggedStringUtf8<tags::GmcpJsonTag>;
 class GmcpMessage final
 {
 private:
-    GmcpMessageName name;
-    std::optional<GmcpJson> json;
```
nschimme, line 66: Why give up on std::optional? Doing an empty check is basically the same thing but it's less obvious to the API user that its actually an optional payload according to the GMCP spec.

src/roompanel/RoomManager.cpp
```diff
@@ -0,0 +1,252 @@
...
+#include <utility>
+#include <QtCore>
+
+#include "../global/StringView.h"
+#include "../proxy/GmcpMessage.h"
+#include "RoomManager.h"
+#include "RoomMobs.h"
+
+RoomManager::RoomManager(QObject *const /*parent*/)
+    : QObject{nullptr}
```
nschimme, line 16: Why not pass in the QMainWindow as the parent? This is will otherwise be leaked when the application is closed.

src/roompanel/RoomManager.cpp
```diff
@@ -0,0 +1,252 @@
...
+    , m_room{std::make_unique<RoomMobs>(nullptr)}
```
nschimme, line 17: Why not make it here? The parent can be the RoomManager ?

src/roompanel/RoomManager.cpp
```diff
@@ -0,0 +1,252 @@
...
+    addMob(doc.object());
+}
+
+void RoomManager::parseGmcpRemove(const GmcpMessage &msg)
+{
+    showGmcp(msg);
+    // payload is a single number (usually followed by a space), not an array or object
+    // thus parsed msg.getDocument() is not valid: use msg.getJson() instead
+    const std::string &str = msg.getJson().getStdString();
+    const unsigned long num = std::stoul(str);
```
nschimme, line 76: Would doing some bound checks here help anything?

src/roompanel/RoomManager.cpp
```diff
@@ -0,0 +1,252 @@
...
+    const RoomMob::Id id = static_cast<RoomMob::Id>(num);
+    if (id == 0 || static_cast<unsigned long>(id) != num) {
+        if (false) {
```
nschimme, line 79: Why disable this warning? Wouldn't we want to discover this error?

src/roompanel/RoomManager.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+void RoomManager::parseGmcpSet(const GmcpMessage &msg)
+{
+    showGmcp(msg);
+    if (isInvalidJson(msg)) {
+        return;
+    }
+    const QJsonDocument &doc = msg.getDocument();
+    if (!doc.isArray()) {
+        if (false) {
```
nschimme, line 99: A member variable (i.e. `debug`) that we can toggle false/true will help us having to manually replace each of these if statements when debugging.

src/roompanel/RoomManager.cpp
```diff
@@ -0,0 +1,252 @@
...
+{
+    QJsonObject::iterator iter = obj.find("id");
+    QJsonObject::iterator end = obj.end();
+    if (iter == end || !toMobId(*iter, data.id())) {
+        return false;
+    }
+    auto matchEnd = mobFields.end();
+    for (iter = obj.begin(); iter != end; ++iter) {
+        auto match = mobFields.find(iter.key());
+        if (match != matchEnd) {
```
nschimme, line 200: What's the impact of this being matchEnd? Should we handle this?

src/roompanel/RoomManager.cpp
```diff
@@ -0,0 +1,252 @@
...
+    return true;
+}
+
+bool RoomManager::toMobId(QJsonValueRef value, RoomMob::Id &id)
+{
+    // RoomMob::Id cannot represent negative numbers
+    const double d = value.toDouble(-1.0);
+    id = static_cast<RoomMob::Id>(d);
+    // check for exact conversion
+    return std::equal_to<double>{}(d, static_cast<double>(id));
```
nschimme, line 213: This isn't very standard... let's pick one of these exceptions: https://www.cplusplus.com/reference/stdexcept/

src/roompanel/RoomManager.cpp
```diff
@@ -0,0 +1,252 @@
...
+    switch (value.type()) {
+    case QJsonValue::Double:
+        variant = static_cast<unsigned>(value.toDouble());
+        break;
+    case QJsonValue::String:
+        str = value.toString();
+        if (str == RoomMobData::you) {
+            str = RoomMobData::YOU;
+        }
+        if (!str.isEmpty() && str.at(0) == QChar('*')) {
```
nschimme, line 230: Please add a comment for why `*`. I imagine this is for `*an Orc*` but it's not super clear.

src/roompanel/RoomWidget.cpp
```diff
@@ -0,0 +1,252 @@
...
+        case ColumnTypeEnum::EFFECTS:
+            return "Effects";
+        case ColumnTypeEnum::WEAPON:
+            return "Weapon";
+        case ColumnTypeEnum::FIGHTING:
+            return "Fighting";
+        case ColumnTypeEnum::MOUNT:
+            return "Mount";
+        default:
+            if (false) {
```
nschimme, line 64: Why not warn here?

src/roompanel/RoomWidget.cpp
```diff
@@ -0,0 +1,252 @@
...
+        return mob->color();
+    }
+    return QColor{};
+}
+
+bool RoomModel::isFightingYOU(const int row, const int column) const
+{
+    if (static_cast<ColumnTypeEnum>(column) == ColumnTypeEnum::FIGHTING //
+        && row >= 0 && size_t(row) < m_mobVector.size()) {
+        //
```
nschimme, line 184: Why these dangling comments?

src/roompanel/RoomWidget.cpp
```diff
@@ -0,0 +1,252 @@
...
+{
+    beginResetModel();
+    m_room.listMobs(m_mobsById, m_mobVector);
+    endResetModel();
+}
+
+// ------------------------------- RoomWidget ----------------------------------
+RoomWidget::RoomWidget(RoomManager *rm, QWidget *const parent)
+    : QWidget{parent}
+    , m_model{nullptr}
```
nschimme, line 201: Don't initialize here then since we're doing it below

src/roompanel/RoomWidget.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+    m_model = std::make_unique<RoomModel>(this, *rm->getRoom());
+
+    m_table = std::make_unique<QTableView>(this);
+    m_table->setSelectionMode(QAbstractItemView::ContiguousSelection);
+    m_table->setSelectionBehavior(QAbstractItemView::SelectRows);
+    m_table->horizontalHeader()->setStretchLastSection(false);
+    m_table->horizontalHeader()->setSectionResizeMode(QHeaderView::Interactive);
+    m_table->setModel(m_model.get());
+    /// TODO: [Cosmos] m_table->setItemDelegate(new RoomDelegate(this));
```
nschimme, line 218: Care to make this comment be more human readable?

src/roompanel/RoomWidget.h
```diff
@@ -0,0 +1,85 @@
...
+signals:
+
+private:
+    void readSettings();
+    void writeSettings();
+
+private:
+    std::unique_ptr<RoomModel> m_model;
+    std::unique_ptr<QTableView> m_table;
+    RoomManager *m_roomManager;
```
nschimme, line 82: Could we pass it in by reference instead?

nschimme (changes requested):

nschimme:
Any chance we could also write some unit tests for this?

nschimme:
@cosmos72 oops, I think I just force pushed over you. Please rebase on top of my change, sorry!

src/pandoragroup/mmapper2group.cpp
cosmos72: The `GmcpMessage` class originally (i.e. before my changes) only contained unparsed JSON inside an UTF-8 string,
and parsed it every time it needed the JSON.

I first changed to it lazily parse the UTF-8 string to JSON on demand, but then I thought about multi-thread synchronization issues - i.e. thread safety - and finally I decided to just parse the UTF-8 string to JSON when GmcpMessage is created.

If you think parsing UTF-8 to JSON is a potentially expensive operation, and it's worth skipping it for messages MMapper is not interested into, I will put back the lazy parsing and its accompanying synchronization.

src/display/mapcanvas.cpp
cosmos72: I could not resist fixing the typo, er, abbreviation :)

src/roompanel/RoomManager.cpp
cosmos72: I'm not an expert of Qt objects lifecycle... I will fix this.

src/roompanel/RoomManager.cpp
cosmos72: Yes, we can ignore strings that contain trail junk or that parse to out-of-range integers. I'll add that.

src/roompanel/RoomWidget.cpp
cosmos72: the `//` are there to force IDEs or clang-format to split these long lines

src/roompanel/RoomWidget.cpp
cosmos72: This is a leftover of copy-pasting `GroupPanel` code. To be honest, I have no idea what's an `ItemDelegate` is. Any reference?

cosmos72:
This branch exists for the pull request, no problem if you pushed commits to it.

src/roompanel/RoomWidget.cpp
nschimme: QStyledItemDelegate allows you to replace the text with some icon: https://doc.qt.io/qt-5/qtwidgets-itemviews-stardelegate-example.html

We used it in the GroupPanel to change the text to little icons

src/global/Flags.h
nschimme: I did this

src/pandoragroup/mmapper2group.cpp
cosmos72: [UPDATE] I modified `GmcpMessage` to lazily parse string to JSON

src/roompanel/RoomWidget.cpp
cosmos72: I see. Removed, not using icons in RoomPanel at the moment.

src/roompanel/RoomWidget.h
cosmos72: Done

src/roompanel/RoomWidget.cpp
cosmos72: Done.

src/roompanel/RoomManager.cpp
cosmos72: Fixed

src/roompanel/RoomManager.cpp
cosmos72: Fixed. I set RoomMobs's parent to `parent` i.e. the main window

src/roompanel/RoomManager.cpp
cosmos72: All JSON object keys are optional in GMCP messages Room.Chars.*  - only 'id' is mandatory.

If `match == matchEnd` it means one of the JSON object keys we know about (name, desc, fighting, flags, labels, riding, driving, position, weapon) is not present in the GMCP message, so we don't save the corresponding value into `RoomMobUpdate`

src/display/mapcanvas.cpp
cosmos72: Reverted. The rename `FALLTHRU` -> `FALLTHROUGH` can be done in a different PR

src/global/Flags.h
cosmos72: shall I revert these constexpr enhancements?

src/roompanel/RoomManager.cpp
cosmos72: Added.

src/roompanel/RoomManager.cpp
cosmos72: Done.

src/roompanel/RoomManager.cpp
cosmos72: I have added
```C++
if (m_debug) {
  qWarning() << // ...
}
```

src/roompanel/RoomManager.cpp
cosmos72: Throw an exception just to immediately catch it in the caller, swallow it and return?
It really seems unnecessarily complicated (and slow) to me.

I wanted to write `return d == static_cast<double>(id);` but some compilers complained that `operator==` on double shouldn't be used.

src/roompanel/RoomManager.cpp
cosmos72: Exactly, it's to highlight enemy players. I added a comment.

src/roompanel/RoomWidget.cpp
cosmos72: I have replaced it with
```C++
if (m_debug) {
  qWarning() << qWarning() << "Unsupported column" << column;
}
```

external/zlib/CMakeLists.txt
```diff
@@ -33,8 +33,8 @@ else()
     message(STATUS "Could not find system zlib; building as a static library")
     include(ExternalProject)
     ExternalProject_Add(zlib
-        URL "https://zlib.net/zlib-1.2.11.tar.gz"
-        URL_HASH SHA256=c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1
+        URL "https://zlib.net/zlib-1.2.12.tar.gz"
+        URL_HASH SHA256=91844808532e5ce316b3c010929493c0244f3d37593afd6de04f71821d5136d9
```
nschimme, line 37: Thanks! This should be a separate PR so we can merge it sooner!

cosmos72: Of course :)
Does it make sense to cherry-pick it for merging in master branch?

cosmos72:
I have rebased 'roompanel' branch - and as a consequence, this PR - on master commit [487b64d0069f3a99e86e04321e106fbe5b0f8c76](https://github.com/MUME/MMapper/commit/487b64d0069f3a99e86e04321e106fbe5b0f8c76)

Hopefully I did not do anything plainly wrong - it's the first time I used `git rebase --interactive` and I also tried to meld together related commits, as per your suggestion

nschimme:
> I have rebased 'roompanel' branch - and as a consequence, this PR - on master commit [487b64d0069f3a99e86e04321e106fbe5b0f8c76](https://github.com/MUME/MMapper/commit/487b64d0069f3a99e86e04321e106fbe5b0f8c76)
> 
> Hopefully I did not do anything plainly wrong - it's the first time I used `git rebase --interactive` and I also tried to meld together related commits, as per your suggestion

I took a stab at organizing the commits a little further.

PR #268, opened 2022-12-27
Fix segfault in _dirs when map has buggy connections.

ethorondil:
Faine's current map has a room connected to a non-existent room, causing a segfault in _dirs.  I don't know how that buggy connection was introduced, but this commit:

 - Fixes the segfault.

 - Adjusts the warning on map load so it's possible to find the bugged room.


nschimme (approved):

nschimme:
Some more thought after I merged it, maybe we should detect this when we load or save the map and prevent the corruption in the first place.

PR #269, opened 2022-12-28
Fix prompt parsing

nschimme:


src/parser/abstractparser.cpp
```diff
@@ -465,6 +465,7 @@ QString AbstractParser::normalizeStringCopy(QString string)
     // transliterations to accidentally count as ANSI.
     ParserUtils::removeAnsiMarksInPlace(string);
     ParserUtils::toAsciiInPlace(string);
+    string.remove(C_CARRIAGE_RETURN);
```
lachupe, line 468: Wont this change any regexp's or lines sent to the user?

nschimme: Possible but I think it's only lines sent for internal parsing. I'll audit it further.

nschimme: I confirmed this doesn't impact what is sent to the user. It looks like for internal actions the logic is unchanged too. Risk is low :-)

PR #271, opened 2022-12-29
Enable hiding both status bar and scroll bars for a more minimal MapWindow

### LINKED ISSUE #251: Option to hide bottom status bar (and possibly scrollbars)
Thank you for the amazing work on this project!

When we record videos for Youtube, we try to compose the map on top of the MUD client. It would be super helpful if we could hide the bottom status bar (the UI element with the season / time of day) and possibly even the scrollbars. They take up a big chunk of real estate when we are trying to compose a screen for Youtube Shorts, etc.

An option under the View... menu for Scrollbars and possibly Toolbars > Bottom Status Bar would be amazing!

mdrcode:
Fixes #251 

Some exploratory and trivial tweaks to enable a menu option under View... for hiding the Status Bar and Scroll Bars. I just followed the general pattern of alwaysOnTop.

One conscious design decision: currently in the MainWindow constructor when it fetches the config value for alwaysOnTop, it inlines the code to manage that setting directly. I thought it made more sense to invoke the slot handler rather than copy-paste code. But eager for guidance here. 

Funny thing on naming: do we call it Scrollbars? ScrollBars? Scroll Bars? Currently in the app we say "Status Bar" but also "Toolbar". @nschimme please decide!

LMK on the debug messages I can delete them.

Also should I squash commits? Forgive my verbose commit style I haven't submitted a PR in 10 years.

### DIFF
--- a/src/display/mapwindow.cpp
+++ b/src/display/mapwindow.cpp
@@ -38,6 +38,8 @@ MapWindow::MapWindow(MapData &mapData, PrespammedPath &pp, Mmapper2Group &gm, QW
 
     m_gridLayout->addWidget(m_horizontalScrollBar.get(), 1, 0, 1, 1);
 
+    m_scrollBarsVisible = true;
+
     m_canvas = std::make_unique<MapCanvas>(mapData, pp, gm, this);
     MapCanvas *const canvas = m_canvas.get();
 
@@ -189,19 +191,29 @@ void MapWindow::slot_setScrollBars(const Coordinate &min, const Coordinate &max)
     updateScrollBars();
 }
 
+void MapWindow::setScrollBarsVisible(const bool visible)
+{
+    qDebug() << "setScrollBarsVisible: " << visible;
+    m_scrollBarsVisible = visible;
+
+    m_horizontalScrollBar->setVisible(visible);
+    m_verticalScrollBar->setVisible(visible);
+
+}
+
 void MapWindow::updateScrollBars()
 {
     const auto dims = m_knownMapSize.size() * MapCanvas::SCROLL_SCALE;
 
     m_horizontalScrollBar->setRange(0, dims.x);
-    if (dims.x > 0) {
+    if (dims.x > 0 && m_scrollBarsVisible) {
         m_horizontalScrollBar->show();
     } else {
         m_horizontalScrollBar->hide();
     }
 
     m_verticalScrollBar->setRange(0, dims.y);
-    if (dims.y > 0) {
+    if (dims.y > 0 && m_scrollBarsVisible) {
         m_verticalScrollBar->show();
     } else {
         m_verticalScrollBar->hide();
--- a/src/display/mapwindow.cpp
+++ b/src/display/mapwindow.cpp
@@ -38,7 +38,7 @@ MapWindow::MapWindow(MapData &mapData, PrespammedPath &pp, Mmapper2Group &gm, QW
 
     m_gridLayout->addWidget(m_horizontalScrollBar.get(), 1, 0, 1, 1);
 
-    m_scrollBarsVisible = std::nullopt;
+    m_showScrollBars = std::nullopt;
 
     m_canvas = std::make_unique<MapCanvas>(mapData, pp, gm, this);
     MapCanvas *const canvas = m_canvas.get();
@@ -191,9 +191,9 @@ void MapWindow::slot_setScrollBars(const Coordinate &min, const Coordinate &max)
     updateScrollBars();
 }
 
-void MapWindow::setScrollBarsVisible(const bool visible)
+void MapWindow::setShowScrollBars(const bool show)
 {
-    m_scrollBarsVisible = visible;
+    m_showScrollBars = show;
     updateScrollBars();
 }
 
@@ -202,14 +202,14 @@ void MapWindow::updateScrollBars()
     const auto dims = m_knownMapSize.size() * MapCanvas::SCROLL_SCALE;
 
     m_horizontalScrollBar->setRange(0, dims.x);
-    if (dims.x > 0 && m_scrollBarsVisible.value_or(true)) {
+    if (dims.x > 0 && m_showScrollBars.value_or(true)) {
         m_horizontalScrollBar->show();
     } else {
         m_horizontalScrollBar->hide();
     }
 
     m_verticalScrollBar->setRange(0, dims.y);
-    if (dims.y > 0 && m_scrollBarsVisible.value_or(true)) {
+    if (dims.y > 0 && m_showScrollBars.value_or(true)) {
         m_verticalScrollBar->show();
     } else {
         m_verticalScrollBar->hide();
--- a/src/display/mapwindow.cpp
+++ b/src/display/mapwindow.cpp
@@ -38,7 +38,7 @@ MapWindow::MapWindow(MapData &mapData, PrespammedPath &pp, Mmapper2Group &gm, QW
 
     m_gridLayout->addWidget(m_horizontalScrollBar.get(), 1, 0, 1, 1);
 
-    m_scrollBarsVisible = true;
+    m_scrollBarsVisible = std::nullopt;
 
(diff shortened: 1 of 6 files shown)


mdrcode:
It looks sexy! 

Screen Shot 2022-12-28 at 8 40 19 PM

Sebster7:
> It looks sexy!
> 
> Screen Shot 2022-12-28 at 8 40 19 PM

How did you get the menu bar to disappear?  I don't see an option for that (in the Windows build) or see you mention or commit a change for that.

Also, a quick way to hide the main toolbar might be useful, rather than having to turn off 9 separate toolbars.
Or a View, Maximal View option (like full-screen, but not full-screen), or something that automatically hides the menu bar, toolbar, status bar and scrollbars?

nschimme:
I usually use browsers as inspiration for UX as a lot of users would be familiar with that:

image

image

I guess Chrome's model is easier for now?

Alternatively we could hide this under Preferences > General > Status Bar and rename that to be more generic.

src/display/mapwindow.cpp
```diff
@@ -189,19 +191,29 @@ void MapWindow::slot_setScrollBars(const Coordinate &min, const Coordinate &max)
     updateScrollBars();
 }
 
+void MapWindow::setScrollBarsVisible(const bool visible)
+{
+    qDebug() << "setScrollBarsVisible: " << visible;
+    m_scrollBarsVisible = visible;
+
+    m_horizontalScrollBar->setVisible(visible);
```
nschimme, line 199: I wonder if `QAbstractScrollArea::setHorizontalScrollBarPolicy( Qt::ScrollBarAlwaysOff )` is better?

https://stackoverflow.com/questions/3383260/how-to-hide-scrollbar-in-qscrollarea

nschimme:
A few more thoughts, we should probably move the clock into a panel if we're going to hide the status bar? It could be a toggle between being a panel or embedded in the status bar.

Also, should we have the status bar still show up if you're hovering over items in the context menus to show the help text? This might require additional coding but would help on the documentation side.

lachupe:
May be just render it in the right low corner of the map, if enabled?

чт, 29 дек. 2022 г., 14:10 Nils Schimmelmann ***@***.***>:

nschimme:
> May be just render it in the right low corner of the map, if enabled?

Yeah, ideally we even render it in GL.

nschimme:
Another thing we could add is showing both the scrollbars and statusbar when a hotkey like Alt is being pressed. (Although perhaps we should just implement our own version of the statusbar at that point?) It seems like this is what browsers are doing if you hover over links.

src/display/mapwindow.cpp
mdrcode: I will investigate this, but according to some discussion on SO it seems this might result in the scroll bar object disappearing completely (i.e. not existing) which could create further issues. Let me take a look and circle back.

mdrcode:
Thank you everyone for the comments! 

@nschimme 
- I certainly don't mean to suggest that we should hide the status bar by default, I was just intending this as a power user option for screen recording, or for someone who consciously wants to stylize a minimal experience.
- Agree, it would be cool to make the status messages float-y or hotkey-able, but yeah as you say I believe that would require us to implement our own status bar since the QT status bar seems to be a "peer" of the map canvas in layout and so every time we toggle it the map widget changes size and stutters/jitters. Perhaps we could overlay the QT widget but that feels a bit heavy weight compared to the more minimal Chrome style.
- If folks seem to like this, I could definitely take a look at having both the clock overlay on the canvas and the status float-y and conditional. 

@Sebster7 
- On Mac, the menu bar is not shown in the app window but lives atop the screen in the system menu bar. I can consider a toggle to hide it on Windows (and maybe Linux?) but then that raises the issue of how would the user resurrect it if they wanted it back?! Context menu? Or maybe just load it visible upon every app start (make it non-sticky, basically, so you have to hide it every time if you want). Needs some thought.
 - I thought about some convenience/helper methods like Hide All Toolbars, but honestly I'm not sure if the convenience is worth the extra code. If many people like a more minimal approach, we can consider something like a global Minimal Mode or something.
  - I also played around with hiding the entire system window frame (so you wouldn't even see the title bar) but that conflicts with the grabbers for resize. It's possible but would need more coding.

mdrcode:
> 



> I guess Chrome's model is easier for now?

Seems a little verbose but good to go (assuming I understood your direction):

Screen Shot 2022-12-29 at 7 27 51 PM

src/mainwindow/mainwindow.cpp
```diff
@@ -309,6 +309,12 @@ MainWindow::MainWindow()
         alwaysOnTopAct->setChecked(true);
         setWindowFlags(windowFlags() | Qt::WindowStaysOnTopHint);
     }
+
+    showStatusBarAct->setChecked(getConfig().general.showStatusBar);
+    slot_setShowStatusBar();
```
mdrcode, line 314: @nschimme Are you ok with calling the slot handler directly? It felt cleaner than duplicating the code to effect the config

nschimme: Yeah, that's fine. They're all functional calls internally if its the same thread anyway :-)

src/display/mapwindow.cpp
```diff
@@ -38,6 +38,8 @@ MapWindow::MapWindow(MapData &mapData, PrespammedPath &pp, Mmapper2Group &gm, QW
 
     m_gridLayout->addWidget(m_horizontalScrollBar.get(), 1, 0, 1, 1);
 
+    m_showScrollBars = std::nullopt;
```
nschimme, line 41: Why not just use `getConfig().general.showScrollBars` here?

nschimme:
> * If folks seem to like this, I could definitely take a look at having both the clock overlay on the canvas and the status float-y and conditional.

Yeah, this isn't really much of a blocker for this feature but I wonder if we would need the status bar hiding at all if we had that. The level of effort is definitely fairly gargantuan for something the average user won't care about.

src/mainwindow/mainwindow.cpp
```diff
@@ -641,10 +647,18 @@ void MainWindow::createActions()
     zoomResetAct->setShortcut(tr("Ctrl+0"));
     zoomResetAct->setStatusTip(tr("Zoom to original size"));
 
-    alwaysOnTopAct = new QAction(tr("Always on top"), this);
+    alwaysOnTopAct = new QAction(tr("Always On Top"), this);
```
nschimme, line 650: Thanks! It's the small details that make it pop 😄

nschimme:
@mdrcode I rebased so the build could go green. I can squash the code as the last merge step, verbosity is good for the review process, thanks!

mdrcode:
@nschimme 
> I rebased so the build could go green

Thank you, I knew I did something wrong there - forgive me as I'm still dusting off all the git cobwebs.

Sebster7:
> @Sebster7
> 
>     * On Mac, the menu bar is not shown in the app window but lives atop the screen in the system menu bar. I can consider a toggle to hide it on Windows (and maybe Linux?) but then that raises the issue of how would the user resurrect it if they wanted it back?! Context menu? Or maybe just load it visible upon every app start (make it non-sticky, basically, so you have to hide it every time if you want). Needs some thought.

Yes, loading it visible upon every app start with an option to hide it, and for it to re-appear when you press Alt would be a decent option and probably not too hard to implement I guess (though I don't know!)

>     * I thought about some convenience/helper methods like Hide All Toolbars, but honestly I'm not sure if the convenience is worth the extra code. If many people like a more minimal approach, we can consider something like a global Minimal Mode or something.

Yes, as I said before in different words, a Minimalist view would be a convenience.
 
>     * I also played around with hiding the entire system window frame (so you wouldn't even see the title bar) but that conflicts with the grabbers for resize. It's possible but would need more coding.

One could just retain an MMapper logo top-left and for that to open up the menu.  Like on MS Office these days.  I realize, and I apologize, that this is a bit off-topic!

PR #273, opened 2022-12-29
Building test version, ignore

lachupe:


src/parser/abstractparser.cpp
```diff
@@ -1514,6 +1515,12 @@ void AbstractParser::genericDoorCommand(QString command, const ExitDirEnum direc
         command = command.replace("$$DOOR$$", cn);
     }
 
+    if (command.startsWith("--[door:")) {
```
nschimme, line 1518: Why this obscure command rather than something more user friendly?

src/timers/CTimers.h
```diff
@@ -0,0 +1,104 @@
+//
+// Created by Azazello on 16.12.22.
+//
+
+#ifndef MMAPPER_CTIMERS_H
```
nschimme, line 5: `#pragma once` is the modern way of doing this pattern :-)

PR #277, opened 2023-01-01
Requesting Discussion: game communication window to show narrates and tells

mdrcode:
Just a crude proof of concept now, but I'm filing a PR to request some discussion and architectural alignment. 

The prevailing pattern we have now is that new windows/sidepanels and features (AutoLogger, RoomManager, and more) are passed all the way into ConnectionListener and ultimately the Proxy class for wiring up the key signals/slots directly from the parsers and telnet clients. I propose a better solution might be:
- Have a single `GameObserver` which watches all game play (both raw text lines sent/received and gmcp as well) and then can emit to listeners.
- All "downstream" features/widgets (like AutoLogger, RoomManager, GameComms (this PR) and more) could register with the Observer and receive signals from it.
- This would simplify the code (and eliminate the verbose constructor chains) and allow us to consolidate common "listener" code, thread proxying, etc.

Concretely, my proposal would be to:
 - Take the current AutoLogger code and refactor that, splitting it out into GameObserver and a new, listening AutoLogger
 - Register my CommsWindow and perhaps the RoomManager (and others) as listeners to the Observer

What do you guys think? If you want to avoid this refactoring now that's totally ok, just felt a bit inelegant to be adding Yet Another Constructor Chained Manager so thought I would ask.

Screen Shot 2022-12-31 at 4 50 16 PM

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -2,6 +2,8 @@ set(mmapper_SRCS
     main.cpp
     adventure/adventurejournal.cpp
     adventure/adventurejournal.h
+    adventure/commswidget.cpp
+    adventure/commswidget.h
     client/ClientTelnet.cpp
     client/ClientTelnet.h
     client/ClientWidget.cpp
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,5 +1,7 @@
 set(mmapper_SRCS
     main.cpp
+    adventure/adventurejournal.cpp
+    adventure/adventurejournal.h
     client/ClientTelnet.cpp
     client/ClientTelnet.h
     client/ClientWidget.cpp
--- a/src/adventure/adventurejournal.cpp
+++ b/src/adventure/adventurejournal.cpp
@@ -1,0 +1,39 @@
+#include "adventurejournal.h"
+#include "global/TextUtils.h"
+#include "parser/parserutils.h"
+
+#include <QDebug>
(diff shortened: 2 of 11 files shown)


nschimme:
This seems fine to me. My only addition is that I've thought about this in the past but worried that we would be losing too much of the state machine in the process if we just emit all of the raw text. There is a lot of state that is captured in the text (such as various XML tags) that would be lost. We might want to consider passing some of that state along the payload.

nschimme:
You might be interested in this class if you haven't noticed it yet: https://github.com/MUME/MMapper/blob/master/src/parser/AbstractParser-Actions.cpp

I thought that we could evaluate certain elements given an XML tag: https://github.com/MUME/MMapper/blob/38eb852032343eb2876f3d45d63227a7819fa2e9/src/parser/mumexmlparser.cpp#L796-L802

nschimme:
... and we do ugly things like this :-)

https://github.com/MUME/MMapper/blob/38eb852032343eb2876f3d45d63227a7819fa2e9/src/parser/mumexmlparser.cpp#L287-L290

nschimme:
Last example, we also register and listen to GMCP events here:

https://github.com/MUME/MMapper/blob/master/src/pandoragroup/mmapper2group.cpp#L557-L593

mdrcode:
Closing out as this is now quite antiquated and the feedback has been consolidated in #281

PR #278, opened 2023-01-03
Game Observer (Refactoring, no-op)

mdrcode:
This should be a functional no-op (no changes from player's perspective).

- As discussed in #277, introduce a `GameObserver` which brokers internal events from `Proxy`, `ConnectListener`, the `Parser`s, and `TelnetClient`s outward to downstream consumers.
- Ideally, after this change, no one should need to touch `ConnectionListener` or `Proxy` unless significantly changing how mapper communicates with the game and external client. Folks wanting to add "mud client style" windows should hopefully be saved that complexity.
- Stop passing `AutoLogger` and 'RoomManager`' instances all the way into `Proxy`.
- In `MainWindow`, port `m_Logger` and `m_roomManager` instances to use a local `m_gameObserver` instance.
- Next up, in future PRs, I will introduce a Comms window (Tells/Narrates) and other game client functionality to leverage GameObserver. 

This turned out to be way simpler than I thought, apologies for the delusions of grandeur on my previous PR.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -200,6 +200,8 @@ set(mmapper_SRCS
     mpi/remoteeditsession.h
     mpi/remoteeditwidget.cpp
     mpi/remoteeditwidget.h
+    observer/gameobserver.cpp
+    observer/gameobserver.h
     opengl/Font.cpp
     opengl/Font.h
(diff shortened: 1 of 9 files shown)


nschimme (approved):
LGTM

PR #281, opened 2023-01-06
Adventure Panel

mdrcode:
Introducing a sneak peek at the Adventure Journal, a new feature for MMapper that helps players track, organize, and understand their adventures. This first release adds a new "side panel" (window) to MMapper which records communication (tells) with other players and XP gained on kills.

Not requesting a formal review or merge just yet, this is alpha quality. There is a tough issue where the XP Gained is sometimes not updated until the next GMCP Char.Vitals message which can take a while. But sharing in the hopes of some feedback. It works (mostly) and I find it useful!

Screen Shot 2023-01-09 at 1 14 00 PM


src/adventure/adventuretracker.cpp
```diff
@@ -0,0 +1,164 @@
...
+    QString str = QString::fromLatin1(ba).trimmed();
+    ParserUtils::removeAnsiMarksInPlace(str);
+
+    // We keep track of the N most recent lines so that we can parse with the
+    // context of a sliding window. Array indices are interpeted backwards,
+    // so [0] is the last line received, [1] is the second-to-last, etc.
+    // Must create our own QString copy for this, since the `ba` passed above
+    // is reused. So shfit the array and add a copy.
+    for (size_t i = m_lastLines.size(); i > 0; i--) {
+        m_lastLines[i] = m_lastLines[i - 1];
```
nschimme, line 37: Aren't we leaking memory here? We're never deleting the QString that was created N + 1 copies ago.

src/adventure/adventuretracker.cpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+    m_lastLines[0] = new QString(str);
+
+    parseIfKillAndXP();
+
+    parseIfAchievedSomething();
+
+    if (str.contains("You gain a level!")) {
```
nschimme, line 45: This might trigger on a whois. Why not do an exact match?

mdrcode: Good catch! I have since updated to match only on the start of a line (so indexOf == 0). But in the mid-term, I should be inferring a level change from GMCP and not need to do this check.

Ultimately, I hope to do as much of this as possible from GMCP (thus significantly mitigating the "whois" issue), and probably the only thing I will parse from text will be the Something is Dead R.I.P. and XP attribution

src/adventure/adventuretracker.cpp
mdrcode: Yes we are leaking. Would you rather I use a list with push/pop or just explicitly delete the N+1 string here?

mdrcode: I've replaced this entirely with stateful parser and no longer maintain a sliding window as a separate copy.

src/adventure/adventuretracker.cpp
```diff
@@ -0,0 +1,164 @@
...
+    if (obj.contains("xp")) {
+        updateXPfromMud(obj["xp"].toDouble());
+    }
+}
+
+void AdventureTracker::parseIfKillAndXP()
+{
+    // We define a "kill with XP earned" event as:
+    //   - The last line received contains "is dead! R.I.P." or "disappears into nothing."
+    //   - AND among the K most recent lines previously received, at least one contains:
```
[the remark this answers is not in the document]
mdrcode: Done, replaced with stateful parsers. No more loopback.

src/adventure/adventuresession.cpp
```diff
@@ -0,0 +1,108 @@
+#include "adventuresession.h"
+#include "QtCore/qdebug.h"
+#include "adventuretracker.h"
+#include "adventurewidget.h"
+#include "xpstatuswidget.h"
+
+AdventureSession::AdventureSession(QString charName)
+    : m_charName{charName}
```
nschimme, line 8: You can do a `: m_charName{std::move(charName)}` here to avoid a copy

src/adventure/adventuresession.cpp
```diff
@@ -0,0 +1,108 @@
...
+    m_xpInitial = src.m_xpInitial;
+    m_xpCheckpoint = src.m_xpCheckpoint;
+    m_xpCurrent = src.m_xpCurrent;
+
+    return *this;
+}
+
+void AdventureSession::endSession()
+{
+    m_isEnded = true;
```
nschimme, line 42: Should we add an assertion to avoid ending a session twice?

src/adventure/adventuresession.cpp
```diff
@@ -0,0 +1,108 @@
...
+    m_endTimePoint = std::chrono::steady_clock::now();
+}
+
+QString AdventureSession::name() const
```
nschimme, line 46: Why not return a const reference?

src/adventure/adventuresession.cpp
```diff
@@ -0,0 +1,108 @@
...
+    auto start = m_startTimePoint;
+    auto end = std::chrono::steady_clock::now();
+    if (m_isEnded)
+        end = m_endTimePoint;
+    auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(end - start);
+
+    double xpAllSession = m_xpCurrent - m_xpInitial;
+    double xpAllSessionPerSecond = xpAllSession / static_cast<double>(elapsed.count());
+
+    return xpAllSessionPerSecond * 3600;
```
nschimme, line 107: Rather than 3600 can we use `std::chrono::hours` ?

src/adventure/adventuresession.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
```
nschimme, line 1: Please add a copyright and license to all the new files

src/adventure/adventuretracker.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+    connect(&m_observer,
+            &GameObserver::sig_sentToUserGmcp,
+            this,
+            &AdventureTracker::slot_onUserGmcp);
+}
+
+void AdventureTracker::slot_onUserText(const QByteArray &ba)
+{
+    QString line = QString::fromLatin1(ba).trimmed();
```
nschimme, line 27: We might be doing this operation in a ton of files. Maybe we need another version of this signal that has it?

src/adventure/adventuretracker.cpp
```diff
@@ -0,0 +1,179 @@
...
+        emit sig_endedSession(*m_session);
+        m_session.reset();
+    }
+}
+
+void AdventureTracker::parseIfReceivedComm(GmcpMessage msg)
+{
+    auto s = msg.getJson()->toQString().toUtf8();
+    QJsonObject obj = QJsonDocument::fromJson(s).object();
+
```
nschimme, line 113: I guess we're at the point where we might be duplicating work here too. Maybe we need a new class for GMCP messages that are parsed. Perhaps an optional member?

src/adventure/adventuretracker.cpp
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+void AdventureTracker::parseIfUpdatedChar(GmcpMessage msg)
+{
+    auto s = msg.getJson()->toQString().toUtf8();
+    QJsonObject obj = QJsonDocument::fromJson(s).object();
+
+    if (!obj.contains("name"))
+        return;
```
nschimme, line 131: We should probably log some warning here?

src/adventure/adventurewidget.cpp
```diff
@@ -0,0 +1,165 @@
...
+    addJournalEntry(msg);
+}
+
+void AdventureWidget::slot_onDied(double xpLost)
+{
+    auto msg = QString(DIED_MSG);
+
+    if (xpLost < 0.0)
+        msg = msg + QString(XP_SUFFIX).arg(formatXPGained(xpLost));
+
```
nschimme, line 96: This logic can be simplified greatly. Let's exit early and only create `msg` once without mutating?

src/adventure/adventurewidget.cpp
```diff
@@ -0,0 +1,165 @@
...
+    auto msg = QString(KILL_TROPHY_MSG).arg(mobName);
+
+    if (xpGained > 0.0) {
+        msg = msg + QString(XP_SUFFIX).arg(formatXPGained(xpGained));
+    } else {
+        // When player gets XP from multiple kills on the same heartbeat, like
+        // frequently happens with quake xp, then the first mob gets all the XP
+        // attributed and the others are zero. No way to solve this given
+        // current "protocol" of how MUME exposes information.
+        msg = msg + QString(XP_SUFFIX).arg("?");
```
nschimme, line 116: MUME's engine event loop works in 250ms pulses so you could assume it's split equally across the current pulse. That's probably better than attributing it to one source (but more code complexity).

src/adventure/adventurewidget.cpp
```diff
@@ -0,0 +1,165 @@
...
+void AdventureWidget::slot_onReceivedHint(const QString &hint)
+{
+    auto msg = QString(HINT_MSG).arg(hint);
+
+    addJournalEntry(msg);
+}
+
+const QString AdventureWidget::formatXPGained(double xpGained)
+{
+    if (abs(xpGained) < 1000) {
```
nschimme, line 131: Do we need to worry about negatives? (i.e. -100k)

src/adventure/adventurewidget.cpp
```diff
@@ -0,0 +1,165 @@
...
+    if (abs(xpGained) < (20 * 1000)) {
+        return QString::number(xpGained / 1000, 'f', 1) + "k";
+    }
+
+    return QString::number(xpGained / 1000, 'f', 0) + "k";
+}
+
+void AdventureWidget::addJournalEntry(const QString &msg)
+{
+    auto prepend = "\n";
```
nschimme, line 144: You could write this as `const auto prepend = (m_numMessagesReceived == 0) ? "" : "\n";`

src/adventure/lineparsers.cpp
```diff
@@ -0,0 +1,99 @@
...
+    return m_lastSuccessVal;
+}
+
+bool AccomplishedTaskParser::parse(QString line)
+{
+    // REVISIT: there are at least three different versions of this
+    //   accomplished
+    //   knowledgeable (xp?)
+    //   travelled (tp?)
+    return line.indexOf("With the task complete, you feel more") == 0;
```
nschimme, line 16: This is technically not super efficient since it checks the whole string for when this begins. The same is probably true for other captures here down the road. A regular expression would know to just check the beginning and abort early (or https://doc.qt.io/qt-5/qstring.html#startsWith with less complexity).

btw, I'm a huge fan of using https://www.regexpal.com/ as a tool to help build regular expressions and test them. We could even do some unit tests to do sanity checks.

src/adventure/lineparsers.h
```diff
@@ -0,0 +1,54 @@
+#pragma once
+
+#include <QString>
+
+class AbstractStatefulLineParser
```
nschimme, line 5: Maybe we need to combine these two? https://github.com/MUME/MMapper/blob/master/src/parser/Action.h

src/adventure/xpstatuswidget.cpp
```diff
@@ -0,0 +1,66 @@
...
+    , m_session{}
+{
+    setFlat(true);
+    setStyleSheet("QPushButton { border: none; outline: none; }");
+
+    setMouseTracking(true);
+
+    updateContent();
+
+    connect(&m_tracker,
```
nschimme, line 19: AdventureTracker can be on a different thread. I wasn't mindful as part of the review around where we might run into race conditions.

src/adventure/lineparsers.cpp
mdrcode: Yeah, d'oh. I guess subconsciously I was thinking that this _was_ `startsWith`, but of course it's not (it's `checkEverywhereUnnecessarily`), will fix.

src/adventure/xpstatuswidget.cpp
mdrcode: Yeah, I've been thinking about this a lot (although not sure those thoughts accomplish much, lol). But in this case (for a UI widget consumer like XPStatus):
- the Tracker only emits a const reference
- the Tracker does not overwrite / rebuild in place when a new session starts, it allocates a new object when a new session starts (previous versions did rebuild in place and I was worried that would cause subtle bugs)
- The UI widget makes a deep copy and only interprets / renders based on the deep copy

Does this seem reasonable?

src/adventure/adventuresession.cpp
mdrcode: Good catch thanks, will fix (an oversight)

src/adventure/adventuretracker.cpp
mdrcode: I struggled with this. IIUC, some of the GMCP messages are encoded objects, some are encoded arrays of objects, and some are just strings ( I think, like `Comm.Channel.Enable`) and some don't have message bodies. So basically we can't just do a blanket .object() across the board. So either we need to:
- Parse the QJsonDocument in a common class, but still let the callers decide what to extract from the document
- Materialize logic for the full GMCP grammar into the class, so that it knows how to parse different message types

Also, there are dragons here: as I commented elsewhere, if you call msg.getJson()->toQString() on a Core.Goodbye message, it crashes (probably because there is no message body and it's seeking endlessly for a terminating character, or something), so I backed off. We probably need a more robust testing strategy for GMCP in general (or at least a strategy for saving a raw message off the wire and then testing it locally).

I didn't feel comfortable diving into the Gmcp code so I felt this little duplication was a reasonable compromise, but let me know.

src/adventure/adventuresession.cpp
mdrcode: Thanks, will do

src/adventure/adventurewidget.cpp
mdrcode: Yeah I am actually tweaking things here to (optionally) include TP as well as XP

src/adventure/adventurewidget.cpp
mdrcode: So the tricky part is: how do I detect the beginning and end of a heartbeat reliably? Blank lines is probably the only protocol, or some combination of blank lines and prompts.

I loaded up several years of MUME logs into a DB, and there is a long tail of obscure scenarios where things can get a little jumbled and lots of messages can appear together, or even mismatched, and tons of corner cases (like your lantern oil runs out in the middle of a fight). The logic I have above (share of experience + R.I.P. within the space of a few lines) is robust against these (in my testing, at least) and covers something like 99.x% of cases. Going after robust, "full heartbeat" parsing is a significant task.

But ultimately, just in terms of personal preference, I'm not sure "smooth average across heartbeat" is that much better. If you're quaking a cave full of snakes or bats, it will probably be more accurate, but if there is any mob level disparity it will still be off. Also, one nice aspect of attributing to one source is that it shows you how much you get for the whole quake at a glance (like "whoa, I just got 11k for quaking that room of snakes").

My strong preference would be to keep the parsing logic simple for initial release and then if anyone actually uses this thing, and likes it, I can consider a more robust heartbeat parsing framework. I worry the parsing complexity is just not worth the risk now

src/adventure/adventuresession.cpp
mdrcode: fixed

src/adventure/adventurewidget.cpp
mdrcode: fixed

src/adventure/lineparsers.cpp
mdrcode: fixed throughout

src/adventure/adventurewidget.cpp
nschimme: Yeah, logs won't have it. The primary chaos factors to be aware of are network lag and the game engine lag. We could probably do a fairly decent estimate but this goes back to the earlier model you had with a sliding window. I agree with your assessment that it's not necessary for the MVP :-)

src/adventure/adventuresession.cpp
```diff
@@ -0,0 +1,126 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (C) 2023 The MMapper Authors
+// Author: Mike Repass <mike.repass@gmail.com> (Taryn)
+
+#include "adventuresession.h"
+#include "QtCore/qdebug.h"
```
nschimme, line 6: why not `#include <QDebug>` ?

src/adventure/adventuretracker.cpp
```diff
@@ -0,0 +1,197 @@
...
+        emit sig_updatedSession(*m_session);
+        return;
+    }
+
+    if (m_session && m_session->name() != charName) {
+        qDebug().noquote() << QString("Adventure: new adventure for %1 replacing %2")
+                                  .arg(charName, m_session->name());
+
+        m_session->endSession();
+        emit sig_endedSession(*m_session);
```
nschimme, line 156: This feels scary because the old `m_session` is destructed after this and whomever is listening might try to keep it around. I wonder if a `std::shared_ptr` would be safer.

src/adventure/adventuretracker.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (C) 2023 The MMapper Authors
+// Author: Mike Repass <mike.repass@gmail.com> (Taryn)
+
+#include "adventure/adventuresession.h"
+#include "adventure/lineparsers.h"
+#include "observer/gameobserver.h"
+#include <fstream>
```
nschimme, line 9: Is `<fstream>` unused in this file?

PR #283, opened 2023-01-13
Add timers command

lachupe:


src/parser/mumexmlparser.cpp
```diff
@@ -172,23 +176,36 @@ void MumeXmlParser::parse(const TelnetData &data, const bool isGoAhead)
...
+{
+    // lvl100 hungry/thirsty?
+    // shall this be simplified somehow?
+    static QRegExp statExp("Needed: * Alert: *.", Qt::CaseSensitive, QRegExp::Wildcard);
+    static QRegExp statExpCond("Needed: * Alert: *. Cond*", Qt::CaseSensitive, QRegExp::Wildcard);
+    static QRegExp lvl100statExp("Gold: *. Alert: *.", Qt::CaseSensitive, QRegExp::Wildcard);
+
+    if (statExp.exactMatch(currentString) || lvl100statExp.exactMatch(currentString)
+        || statExpCond.exactMatch(currentString)) {
+        sendToUser(m_timers.getStatCommandEntry(), isGoAhead);
```
nschimme, line 208: isGoAhead is only for prompts so we don't need to pass this argument through. Given this, we can also keep this logic confined to parseMudCommands()

src/timers/CTimers.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+CTimers::CTimers(QObject *parent)
+    : QObject(parent)
+{
+    m_nextId = 1;
+}
+
+CTimers::~CTimers() {}
+
+void CTimers::addTimer(QByteArray name, QByteArray desc)
```
nschimme, line 18: We can avoid making copies of the arguments here.

src/timers/CTimers.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+    // append the shot-down event pending
+    QTimer::singleShot(time, this, SLOT(finishCountdownTimer()));
+
+    m_countdowns.push_back(countdown);
+}
+
+void CTimers::finishCountdownTimer()
+{
+    QMutexLocker locker(&m_lock);
```
nschimme, line 79: I'm wondering if a mutex is even necessary here. Isn't the Thread owned by the same object as the Proxy and Parser so it's the same thread?

src/timers/CTimers.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+    int deleted = 0;
```
nschimme, line 81: Does it matter how many we delete? Can this be a boolean?

src/proxy/proxy.cpp
```diff
@@ -121,10 +121,12 @@ void Proxy::slot_start()
     m_mudTelnet = makeQPointer<MudTelnet>(this);
     m_telnetFilter = makeQPointer<TelnetFilter>(this);
     m_mpiFilter = makeQPointer<MpiFilter>(this);
+    m_timers = makeQPointer<CTimers>(this);
```
nschimme, line 124: Why do we create this object in the Proxy as opposed to in the AbstractParser?

src/parser/mumexmlparser.cpp
```diff
@@ -160,23 +160,34 @@ void MumeXmlParser::parse(const TelnetData &data, const bool isGoAhead)
...
-                && tempStr.at(1) == m_snoopChar.value() && tempStr.at(2) == ' ') {
-                // Remove snoop prefix (i.e. "&J Exits: north.")
-                tempStr = tempStr.mid(3);
-            }
-            parseMudCommands(tempStr);
-        }
+void MumeXmlParser::printTimersExtensions(const QString currentString, bool isGoAhead) {
+    // lvl100 hungry/thirsty?
+    // shall this be simplified somehow?
+    static QRegExp statExp("Needed: * Alert: *.", Qt::CaseSensitive, QRegExp::Wildcard);
```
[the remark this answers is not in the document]
lachupe: Nice, I'll try that one!

src/parser/mumexmlparser.cpp
lachupe: Not yet sure I understood the hint about keeping the logic in parseMudCommands (need to take a look at the code). But I guess i can remove the isGoAhead. This is for sure only about injecting additional lines right after the matching one.

src/timers/CTimers.cpp
lachupe: Well, I wasn't sure. 
At least 3 things may happen (in theory) at the same time:
- timer for the countdown kicking in and tweaking the countdowns data
- User input adds new entries or remove the old ones
- Proxy (if i recall that right) attempting to retrieve the entries

Now, I'm not sure if this is really in the same thread or not, so I kept the same protection as the old code in PandoraMapper.

src/timers/CTimers.cpp
lachupe: Probably I'd better opt for writing out how many items were deleted. timers may have the same "name", so multiples may be deleted.

src/timers/CTimers.cpp
```diff
@@ -0,0 +1,152 @@
...
+{
+    if (m_timers.size() == 0)
+        return "";
+
+    QString line = "Timers:\r\n";
+    for (std::list<TTimer>::iterator it = m_timers.begin(); it != m_timers.end(); ++it) {
+        line += QString("- %1 < %2 > (up for - %3)\r\n")
+                    .arg(it->name.data(), it->desc.data(), msToMinSec(it->timer.elapsed()));
+    }
+    return line.toLocal8Bit();
```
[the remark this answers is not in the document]
lachupe: Sounds good to me. I don't think there is any solid reason (matching and regexps may be?) to keep QStrings and QByteArray. This is also only due to historical reasons of the code coming from pandora.

src/timers/CTimers.cpp
lachupe: :+1:

src/timers/CTimers.h
```diff
@@ -0,0 +1,66 @@
...
+#include <QMutex>
+#include <QObject>
+
+struct TTimer
+{
+    int id;
+    QByteArray name;
+    QByteArray desc;
+    int duration;
+    QElapsedTimer timer;
```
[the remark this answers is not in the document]
nschimme: I couldn't think of one. I was wondering if we need to track N different QTimers here of we just had a "master" one somewhere that would parse the data and learn when to run the next callback. This thing here is more of a thought experiment than a real need so feel free to ignore it.

src/parser/mumexmlparser.cpp
nschimme: Ah, let me give the backstory. The general Parser file is getting huge so I'm trying to isolate similar logic to other files

src/parser/mumexmlparser.cpp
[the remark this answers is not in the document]
lachupe: Works! Regexp magic! ;)

src/timers/CTimers.cpp
lachupe: I switched to std::strings and using const references. 
Saw one issue with latin-1 encoding, but that got to be my conversions somewhere...

src/timers/CTimers.cpp
lachupe: Yeah, replaced with a boolean. 
Int wasn't necessary at all.

src/timers/CTimers.cpp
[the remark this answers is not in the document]
lachupe: Replaced! Checking for bugs.

src/parser/mumexmlparser.cpp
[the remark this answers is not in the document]
nschimme: I just realized it won't work for ANSI colors. I'll have to give you and updated regex.

src/timers/CTimers.cpp
```diff
@@ -0,0 +1,172 @@
...
+        } else {
+            ++it;
+        }
+    }
+
+    // but still, at least 1ms
+    delta = std::max(delta, 1);
+
+    // as of Qt 5.2.1 singleShot timers are not precise enough and might, easily, fire earlier!
+    // Aza: check? I have no idea what this is about anymore
```
nschimme, line 123: Should we re-check rather than carry this comment forward?

src/parser/mumexmlparser.h
```diff
@@ -99,6 +100,7 @@ class MumeXmlParser final : public AbstractParser
     NODISCARD bool element(const QByteArray &);
     void move();
     NODISCARD std::string snoopToUser(const std::string_view str);
+    void printTimersExtensions(const QString currentString);
```
nschimme, line 103: can pass by const reference

src/CMakeLists.txt
```diff
@@ -386,6 +387,8 @@ set(mmapper_SRCS
     syntax/User.h
     syntax/Value.cpp
     syntax/Value.h
+    timers/CTimers.cpp
```
nschimme, line 390: Can we remove the C prefix? It doesn't match the style of the rest of the files

src/parser/mumexmlparser.cpp
```diff
@@ -172,23 +176,31 @@ void MumeXmlParser::parse(const TelnetData &data, const bool isGoAhead)
...
-            if (m_snoopChar.has_value() && tempStr.length() > 3 && tempStr.at(0) == '&'
-                && tempStr.at(1) == m_snoopChar.value() && tempStr.at(2) == ' ') {
-                // Remove snoop prefix (i.e. "&J Exits: north.")
-                tempStr = tempStr.mid(3);
-            }
-            parseMudCommands(tempStr);
-        }
+void MumeXmlParser::printTimersExtensions(const QString currentString)
+{
+    static QRegExp statLineExp("^(?:Needed:(?: [\\d,]+ xp)?(?:,? [\\d,]+ tp)\\. )?Gold: [\\d,]+\\.(?: Iv: [^.]+\\.)? Alert: \\w+\\.(?: Condition: [^.]+\\.)?");
```
nschimme, line 200: Oh, `QRegExp` is a deprecated class in Qt. Let's switch to `QRegularExpression`

src/timers/CTimers.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+void CTimers::addCountdown(const std::string &name, const std::string &desc, int time)
+{
+    QMutexLocker locker(&m_lock);
+
+    TTimer countdown;
+    countdown.name = name;
+    countdown.desc = desc;
+    countdown.timer.start();
+    countdown.id = m_nextId++;
```
nschimme, line 84: Do we even need the `id` ?

src/timers/CTimers.cpp
nschimme: I left it

src/timers/CTimers.h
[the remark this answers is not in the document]
nschimme: I found a simpler solution, see the refactor

src/parser/mumexmlparser.cpp
```diff
@@ -170,23 +174,21 @@ void MumeXmlParser::parse(const TelnetData &data, const bool isGoAhead)
         m_tempCharacters.clear();
     }
     if (!m_lineToUser.isEmpty()) {
-        sendToUser(m_lineToUser, isGoAhead);
```
nschimme, line 173: This line might cause a regression

PR #294, opened 2023-02-12
Added warrens dawn/dusk messages

mvancleave99:
Added a couple of dawn/dusk messages that the game sends when a player is around warrens.

### DIFF
--- a/src/clock/mumeclock.cpp
+++ b/src/clock/mumeclock.cpp
@@ -88,7 +88,7 @@ const QHash<QString, MumeTimeEnum> MumeClock::m_stringTimeHash{
     // Outside Warrens
     {"Light gradually filters in, proclaiming a new sunrise.", MumeTimeEnum::DAWN},
     {"The deepening gloom announces another sunset.", MumeTimeEnum::DUSK},
-    {"The Rising sun illuminates the landscape.", MumeTimeEnum::DAWN},
+    {"The rising sun illuminates the landscape.", MumeTimeEnum::DAWN},
 };
 
 MumeClock::MumeClock(int64_t mumeEpoch, QObject *parent)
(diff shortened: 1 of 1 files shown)


src/clock/mumeclock.cpp
```diff
@@ -84,6 +84,11 @@ const QHash<QString, MumeTimeEnum> MumeClock::m_stringTimeHash{
     {"The sun slowly climbs from the vale of Anduin in the east.", MumeTimeEnum::DAWN},
     {"Shadows surround you as the sun disappears behind the peaks of the Hithaeglir.",
      MumeTimeEnum::DUSK},
+    
+    // Outside Warrens
+    {"Light gradually filters in, proclaiming a new sunrise.", MumeTimeEnum::DAWN},
+    {"The deepening gloom announces another sunset.", MumeTimeEnum::DUSK},
+    {"The Rising sun illuminates the landscape.", MumeTimeEnum::DAWN},
```
nschimme, line 91: Typo? Capital R in rising?

mvancleave99: Correct. That is a typo. I'll fix.

mvancleave99: Fixed and recommitted.

mvancleave99:
Is there any reason not to use regular expressions for these? That would reduce the number of entries and possibly capture some additional missing regional message.

nschimme:
> Is there any reason not to use regular expressions for these? That would reduce the number of entries and possibly capture some additional missing regional message.

Nobody has coded support for it yet? :-)

I have since added some type of Regex support in this class but I had not get around to combining the two mechanisms: https://github.com/MUME/MMapper/blob/master/src/parser/AbstractParser-Actions.cpp

Please feel free to enhance this yourself!

mvancleave99:
> 

Yeah, I may. That's sort of why I asked. I may need help with getting the C++ part right, but can certainly develop the regex.

mvancleave99:
> 

Nope. Looking at the code you linked, this is way beyond my capability.

Issue #306, opened 2023-03-27
Revealing the Map as you Explore (starts hidden by default)


Sebster7:
I think it would be really cool and exciting for newbies if the default MMapper map started essentially blank with just the room you login to, and was revealed as you explored the terrain (think Sid Meier's Civilisation or Sid Meier's Colonization or many other games).  

How this would work:
1. Rooms get a new property, 'visited' or 'revealed' or 'explored' or some similar meaning name, datetime, default [].
2. All rooms with this new property = [] (i.e. not set) are hidden on the map.
3. All rooms where this new property is true become unhidden on the map, with their exit flags, mob flags, obj flags, etc. becoming visible.
4. When you enter a room (and MMapper knows where you are), or scout a room, the new property gets set to the current real-world time (probably based on your local computer time), meaning it now evaluates to true and the room is revealed as per number 3.
5. Exit keywords/flags for unrevealed rooms would not be shown.

Why?
1. Why a timestamp rather than just using True?  Well, as a future possible feature, we could add the expanding map to a time-lapsed video journal that can be viewed in MMapper or exported and shared with friends.  This journal could also include @mdrcode 's new adventure journal log, so it shows the expanding map and the character's progress (think Sid Meier's Railroad Tycoon, and many other strategy games).
2. We shouldn't be giving newbies the keys to the kingdom before they even log in for the first time.  They should have to explore for themselves, and if they don't, they are missing out on a lot of the excitement and magic of MUME.
3. Revealing the Map as they go shows them their progress and achievement in exploring the MUME world, encouraging them to explore more.
4. But this is better than not sharing a map with them at all, because (a) creating a new map is an effort they may not want to go to, and (b) some 'helpful' person will just tell them to download F****'s map, without realising this is spoiling the newbie's enjoyment.

Optional extras:
1. a button to temporarily reveal the whole map, but only while you have the button held down with the mouse!  (So it's not particularly practical to 'cheat', but you still can if you want to.  This also allows someone to take screenshots of the entire map.)
2. Instead of a simple property: an array with {RevealedBy: CharacterName string; RevealedDT: timestamp datetime}, i.e. MMapper would get your character name from GMCP, and use that and update the map with both the character name and the revealed time of the room.  This would allow for the adventure journal to have properly separate time-lapsed video maps per character, as well as separate progress per character.  It could also be used to avoid the export map / hack property to true and re-open map exploit described below for most people - they would have to do it for each character they play.
[...]

PR #308, opened 2023-04-03
Initial support for MUME's room id

cosmos72:
Parses MUME's room id into a newly created class RoomServerId,
and also adds a RoomServerId field to relevant classes: Room and ParseEvent.

Definitely NOT complete, but it's a first step.

The path machine code definitely needs review from someone who knows it better than me.


cosmos72:
I'll try to answer while waiting for @nschimme take on this:
This patch only contains the bare minimum needed to receive, extract and pass around MUME room ids.

The next step is extending path machine algorithms to add MUME room ids as a very high priority/confidence source of location info.
Unluckily, the path machine is definitely complicated and I currently don't have the time to study it - someone else (@nschimme ?) will be needed to extend it.

Sebster7:
Thanks for your reply @cosmos72.  I'd say that it would be good to get in the parsing and saving of the **current** room id to the map, and ignore everything else (as that may - or should - yet change).  This way, people can start to get their maps updated (as the default map having this is sort of a blocker on #306 ).

cosmos72:
Hi @Sebste7,
that was exactly my thought when I started adding support for room ids.

Although, it turned out that this means reconciling room ids with the other mechanisms used by path machine for recognizing and identifying rooms.

So you still have to modify the path machine, although the changes are probably simpler: after all, the intention is to only WRITE room ids into room, not to READ them back and use them as a source of truth inside path machine algorithm.

nschimme:
Merged as part of db0806068d72f81af8bc26df9607a250bf066ed1

PR #309, opened 2023-04-05
Fix unit tests

nschimme:


tests/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ set_target_properties(
   CXX_STANDARD_REQUIRED ON
   CXX_EXTENSIONS OFF
   COMPILE_FLAGS "${WARNING_FLAGS}"
-  UNITY_BUILD ON
+  UNITY_BUILD ${USE_UNITY_BUILD}
```
mdrcode, line 23: Just curious - is there no way to set all these target_properties in a common location and then inherit? Not a huge deal of course just learning.

tests/CMakeLists.txt
```diff
@@ -189,13 +189,14 @@ set(adventure_SRCS
         )
 set(TestAdventure_SRCS testadventure.cpp testadventure.h)
 add_executable(TestAdventure ${TestAdventure_SRCS} ${adventure_SRCS})
+add_dependencies(TestAdventure glm)
```
mdrcode, line 192: So just to double check, *all* tests should add this glm dependency explicitly?

mdrcode (approved):
LGTM

tests/CMakeLists.txt
nschimme: GmcpMessage seems to be dependent on src/global/bits.h

We should probably convert our core business logic into a lib that is used by the executable and tests

tests/CMakeLists.txt
nschimme: I dug through the documentation and couldn't find a way to set it globally

PR #310, opened 2023-04-07
use GMCP Event.Darkness, Event.Moon, Event.Sun to sync time

nschimme:


src/clock/mumeclock.cpp
```diff
@@ -199,92 +149,100 @@ void MumeClock::parseMumeTime(const QString &mumeTime, const int64_t secsSinceEp
...
         break;
     case MumeTimeEnum::DUSK:
         moment.hour = dusk;
+        type = "sunset";
         break;
     case MumeTimeEnum::NIGHT:
         moment.hour = dusk + 1;
+        type = "night";
         break;
     case MumeTimeEnum::UNKNOWN:
```
mdrcode, line 230: Would it be better to just add WEATHER as a formal enum value?

src/proxy/proxy.cpp
```diff
@@ -169,6 +169,7 @@ void Proxy::slot_start()
             &m_groupManager,
             &Mmapper2Group::slot_parseGmcpInput);
     connect(mudTelnet, &MudTelnet::sig_relayGmcp, m_parserXml, &MumeXmlParser::slot_parseGmcpInput);
+    connect(mudTelnet, &MudTelnet::sig_relayGmcp, &m_mumeClock, &MumeClock::slot_parseGmcpInput);
```
mdrcode, line 172: At some point I would personally suggest that we extract MumeClock from the proxy wiring and have it live outside, hooking up only to GameObserver. But not a blocker here

src/clock/mumeclock.cpp
```diff
@@ -41,57 +41,7 @@ const QMetaEnum MumeClock::s_westronWeekDayNames
 const QMetaEnum MumeClock::s_sindarinWeekDayNames
     = QMetaEnum::fromType<MumeClock::SindarinWeekDayNamesEnum>();
 
-const QHash<QString, MumeTimeEnum> MumeClock::m_stringTimeHash{
```
mdrcode, line 44: So the big takeaway here is that the game now sends concrete time-of-day events, which eliminates the need for hardcoding zone-specific string "actions"?

src/clock/mumeclock.cpp
```diff
@@ -199,92 +149,100 @@ void MumeClock::parseMumeTime(const QString &mumeTime, const int64_t secsSinceEp
...
     // Predict current hour given the month
     const auto dawnDusk = getDawnDusk(moment.month);
     const int dawn = dawnDusk.dawnHour;
     const int dusk = dawnDusk.duskHour;
+
+    const char *type;
     switch (time) {
     case MumeTimeEnum::DAWN:
         moment.hour = dawn;
+        type = "sunrise";
```
mdrcode, line 216: type is only used for debug logging? nothing else?

src/clock/mumeclock.cpp
```diff
@@ -199,92 +149,100 @@ void MumeClock::parseMumeTime(const QString &mumeTime, const int64_t secsSinceEp
...
+            time = MumeTimeEnum::NIGHT;
+            break;
+        case 'r': // rise
+            time = MumeTimeEnum::DAWN;
+            break;
+        case 's': // set
+            time = MumeTimeEnum::DUSK;
+            break;
+        default:
+            assert(false);
```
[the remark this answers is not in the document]
nschimme: It will throw an assertion (aka crash) during a debug build only. We want this by design, rather than it silently failing. I suppose there is no harm in adding a warning but should be pretty clear from the core dump or assertion point.

src/clock/mumeclock.cpp
nschimme: Weather was the previous string in the debug statement (will continue answer in next comment)

src/proxy/proxy.cpp
nschimme: Yeah, I wasn't sure on the best pattern here. I will pass the observer in and then connect it in the clock like what you did with the adventure panel. It is probably better for testing anyway since it's more isolated.

src/clock/mumeclock.cpp
nschimme: Continuing previous comment, yes, it's only used for debug

src/clock/mumeclock.cpp
nschimme: I renamed this to "reason" to better identify it as a debugging feature

nschimme:
@mdrcode I believe I addressed all your feedback

mdrcode (approved):
LGTM

mdrcode (approved):

Issue #312, opened 2023-04-14
Moon cycle change 2704 breaks MMapper2 moon counter


Kaspar-Metsa:
Dain changed the moon cycle: https://mume.org/news/game/2704.php

The new moon cycle formula:
#define MOON_CYCLE_MINS = ((29 * MUD_HOURS_PER_DAY + 12) * MUD_MINS_PER_HOUR + 44)

It's the real world length of a synodic month rounded to minutes:
https://en.wikipedia.org/wiki/Lunar_month#Synodic_month

This issue is to fix existing MMapper2 moon cycle functionality. When working on this please also consider some moon-related ideas mentioned here: https://github.com/MUME/MMapper/issues/280

kasparAxual:
More information regarding new moon calculations that Dain (very) kindly provided:

*The moon's **position in the sky** (e.g., "moonrises in the east") and its **phase** (whether it's new/half/full) are completely different topics.*

**Dain**
> "First, the earth rotates one lap (taking... 24 hours). During that time, the moon has moved a bit. The earth needs to 
> catch up to that delta so it takes (MOON_CYCLE × 1d) ÷ (MOON_CYCLE - 1d) in total."

**Question**
I am sorry but can you bring an example how to use this formula please? What is "1d" and why are we multiplying MOON_CYCLE with "1d"?

**Dain**:
> 
> 1d is one (MUME) day.
> 
> Let's say you want the moon & earth to return to the same relative position (for example "moon rising in
> the east"). Assume they are in that position.
> 
> After additional time _t_, the earth has spun _t_ ÷ 1day revolutions. In the same time, the moon has spun
> _t_ ÷ 29d12h44min.
> 
> If you want their relative position to return to the same state, the earth needs to spin one revolution
> plus some fraction _f_, and that fraction has to equal how far the moon spun.
> 
>   f = t ÷ 1day - 1 = t ÷ 29d12h44min
> 
> Multiply both sides by (1day × 29d12h44min) and you get
> 
>   t × 29d12h44min - 1day × 29d12h44min = t × 1day
> 
> Move things around:
> 
>   t × 29d12h44min - t × 1day = 1day × 29d12h44min
> 
> Break out common factor:
> 
>   t × (29d12h44min - 1day) = 1day × 29d12h44min
> 
> and divide by (29d12h44min - 1day):
> 
>   t = (1day × 29d12h44min) ÷ (29d12h44min - 1day)

kasparAxual:
**Question**
[...]

PR #313, opened 2023-04-22
Moon follows a synodic month

nschimme:
See https://github.com/MUME/MMapper/issues/312

### DIFF
--- a/src/clock/mumeclockwidget.cpp
+++ b/src/clock/mumeclockwidget.cpp
@@ -63,7 +63,7 @@ void MumeClockWidget::slot_updateLabel()
     const MumeClockPrecisionEnum precision = m_clock->getPrecision();
 
     bool updateMoonText = false;
-    const MumeMoonPhaseEnum phase = moment.toMoonPhase();
+    const MumeMoonPhaseEnum phase = moment.moonPhase();
     if (phase != m_lastPhase) {
         m_lastPhase = phase;
         switch (phase) {
@@ -160,10 +160,10 @@ void MumeClockWidget::slot_updateLabel()
     } else
         timeLabel->setText(m_clock->toCountdown(moment));
 
-    const MumeMoonVisibilityEnum moonVisibility = moment.toMoonVisibility();
+    const MumeMoonPositionEnum moonVisibility = moment.moonPosition();
     if (moonVisibility != m_lastVisibility || updateMoonStyleSheet) {
         m_lastVisibility = moonVisibility;
-        const QString moonStyleSheet = (moonVisibility == MumeMoonVisibilityEnum::HIDDEN)
+        const QString moonStyleSheet = (!moment.isMoonVisible())
                                            ? "color:black;background:grey"
                                            : ((moment.isMoonBright() && time >= MumeTimeEnum::DUSK)
                                                   ? "color:black;background:yellow"
--- a/src/clock/mumemoment.cpp
+++ b/src/clock/mumemoment.cpp
@@ -6,24 +6,8 @@
 
 #include <iostream>
 
-#include "../global/Array.h"
 #include "../global/Debug.h"
 #include "mumeclock.h"
-
-static constexpr const int MUME_MINUTES_PER_HOUR = 60;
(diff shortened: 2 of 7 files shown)


src/clock/mumeclockwidget.cpp
```diff
@@ -160,14 +160,14 @@ void MumeClockWidget::slot_updateLabel()
     } else
         timeLabel->setText(m_clock->toCountdown(moment));
 
-    const MumeMoonVisibilityEnum moonVisibility = moment.toMoonVisibility();
+    const MumeMoonVisibilityEnum moonVisibility = moment.moonVisibility();
     if (moonVisibility != m_lastVisibility || updateMoonStyleSheet) {
         m_lastVisibility = moonVisibility;
-        const QString moonStyleSheet = (moonVisibility == MumeMoonVisibilityEnum::HIDDEN)
+        const QString moonStyleSheet = (moonVisibility <= MumeMoonVisibilityEnum::INVISIBLE)
```
mdrcode, line 166: This <= feels odd to me from a readability standpoint. Is there a simpler way to do this?

mdrcode (dismissed):
LGTM. I read through this for looking for legibility and maintainability more than a deep understanding of the crazy moon logic. It is very impressive dedication to code all this!

src/clock/mumeclockwidget.cpp
nschimme: I'll just do an or instead :-)

nschimme:
@mdrcode please review again if you have the time

src/clock/mumemoment.cpp
```diff
@@ -146,133 +130,175 @@ MumeTimeEnum MumeMoment::toTimeOfDay() const
...
 }
 
-int MumeMoment::moonSet() const
+int MumeMoment::moonLevel() const
 {
-    return (moonCycle() + 18) % MUME_DAYS_PER_MOON_CYCLE;
+    // Illumination level seems to be between 0-12
+    // 0 is new moon, 12 is full moon
+    // Levels > 4 make the moon show up in the prompt as ')'
+    int level = (moonZenithMinutes() + MUME_MINUTES_PER_HOUR / 2) / MUME_MINUTES_PER_HOUR;
```
mdrcode, line 165: Hmmm, this is a little cryptic to me - why add MUME_MINUTES_PER_HOUR /2 in the first "term" here? What does that do?

src/clock/mumemoment.cpp
```diff
@@ -146,133 +130,175 @@ MumeTimeEnum MumeMoment::toTimeOfDay() const
...
-    // Levels > 4 make the moon show up in the prompt as ')'
-    const int dayOfCycle = moonCycle();
-    if (dayOfCycle <= 12)
-        return dayOfCycle; // Increasing illumination
-    else
-        return MUME_DAYS_PER_MOON_CYCLE - dayOfCycle; // Decreasing illumination
+    if (isMoonBelowHorizon() || moonPhase() == MumeMoonPhaseEnum::NEW_MOON)
+        return MumeMoonVisibilityEnum::INVISIBLE;
+
+    const auto isBright = isMoonBright();
```
mdrcode, line 192: I personally would prefer just renaming the members to be a bit shorter over this aliasing idiom, but nbd either way

mdrcode (approved):
Impressive dedication. Given the complexity, imho the biggest "requirement" isn't so much getting this perfect but having very graceful failure mode?

src/clock/mumemoment.cpp
nschimme: Yeah, I'm a little tired of staring at this code and just want to get it out there to users. It's more important at this point that it works rather than it being perfectly maintainable. All it takes is Dain changing it once more... :-P

src/clock/mumemoment.cpp
nschimme: To be honest, this is just what I could get working. Cryptic is the name of reverse engineering and in this case I tried re-using some other mechanisms but it wouldn't be as stable in synchronizing.

PR #340, opened 2024-02-02
Added Mud Server Status Protocol

Mattias-Viklund:
Handles game time.

(Amend) Removed CMakePresets/CMakeUserPresets.

### DIFF
--- a/src/clock/mumeclock.cpp
+++ b/src/clock/mumeclock.cpp
@@ -291,6 +291,31 @@ void MumeClock::parseClockTime(const QString &clockTime, const int64_t secsSince
     m_mumeStartEpoch = newStartEpoch;
 }
 
+void MumeClock::parseMSSP(const int year, const int month, const int day, const int hour)
+{
+    // We should not parse the fuzzy MSSP time if we already have a greater precision.
+    if (getPrecision() > MumeClockPrecisionEnum::DAY)
+        return;
+
+    const int64_t secsSinceEpoch = QDateTime::QDateTime::currentDateTimeUtc().toSecsSinceEpoch();
+
+    auto moment = getMumeMoment();
+    moment.year = year;
+    moment.month = month;
+    moment.day = day;
+    moment.hour = hour;
+    // Don't override minute, since we don't get the from the MSSP time.
+
+    const int64_t newStartEpoch = secsSinceEpoch - moment.toSeconds();
+    m_mumeStartEpoch = newStartEpoch;
+
+    // Update last sync timestamp
+    setLastSyncEpoch(secsSinceEpoch);
+
+    setPrecision(MumeClockPrecisionEnum::HOUR);
+    log("Synchronized clock using MSSP");
+}
+
 // TODO: move this somewhere useful?
 NODISCARD static const char *getOrdinalSuffix(const int day)
 {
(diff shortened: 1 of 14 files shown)


src/proxy/MudTelnet.cpp
```diff
@@ -215,3 +221,84 @@ void MudTelnet::sendCoreSupports()
 
     sendGmcpMessage(GmcpMessage(GmcpMessageTypeEnum::CORE_SUPPORTS_SET, set));
 }
+
+void MudTelnet::parseMudServerStatus(const QByteArray &data)
+{
+    QMap<std::string, std::vector<std::string>> map;
```
nschimme, line 227: Can we use std::map instead?

src/proxy/MudTelnet.cpp
```diff
@@ -215,3 +221,84 @@ void MudTelnet::sendCoreSupports()
...
+
+    std::string varName;
```
nschimme, line 229: Let's wrap this in std::<optional> as it should be cleared when a value is set?

src/proxy/proxy.h
```diff
@@ -7,6 +7,7 @@
 
 #include <memory>
 #include <QByteArray>
+#include <QMap>
```
nschimme, line 10: Unnecessary include?

src/proxy/proxy.cpp
```diff
@@ -565,3 +574,14 @@ bool Proxy::isGmcpModuleEnabled(const GmcpModuleTypeEnum &module) const
...
+
+void Proxy::slot_onSendGameTimeToClock(const int year,
+                                       const std::string &monthStr,
+                                       const int day,
+                                       const int hour)
+{
+    const int month = m_mumeClock.getMumeMonth(::toQStringLatin1(monthStr));
+    MumeMoment moment(year, month, day, hour, 0);
+    QString mumeTime = m_mumeClock.toMumeTime(moment);
+    m_mumeClock.parseMumeTime(mumeTime);
```
nschimme, line 586: Doesn't this adjust the granularity of the clock? We're not aware of the seconds marker so we should keep it fuzzy.

src/proxy/MudTelnet.cpp
```diff
@@ -215,3 +221,84 @@ void MudTelnet::sendCoreSupports()
...
+            }
+            lastType = TNSB_MSSP_VAL;
+            continue;
+        }
+        charBuf.append(data.at(i)); // Regular character.
+    }
+
+    //const std::map<std::string, std::vector<std::string>> cMap(map);
+    auto firstElement([](std::vector<std::string> elements) {
+        if (elements.size() == 0)
```
nschimme, line 283: empty?

src/clock/mumeclock.cpp
```diff
@@ -410,3 +410,12 @@ MumeClock::DawnDusk MumeClock::getDawnDusk(const int month)
     const auto m = static_cast<uint32_t>(month);
     return DawnDusk{s_dawnHour.at(m), s_duskHour.at(m)};
 }
+
+int MumeClock::getMumeMonth(const QString &monthName)
```
nschimme, line 414: Should we have this return the enumerated value instead?

src/proxy/proxy.cpp
Mattias-Viklund: I didn't think about that, I'll change it.

src/proxy/MudTelnet.cpp
Mattias-Viklund: Yeah of course, didn't know it existed.

src/clock/mumeclock.cpp
Mattias-Viklund: I'll have a look at it.

src/proxy/proxy.cpp
Mattias-Viklund: So 
`m_mumeClock.setPrecision(MumeClockPrecisionEnum::DAY);`
?

src/clock/mumeclock.cpp
nschimme: This might be changing the entire data structure and a bunch of other code. If so, it's probably best saved for a future PR.

src/proxy/proxy.cpp
nschimme: Well, you're basically taking a MumeTime and then changing it to a String and then parsing that String. I would avoid the string transformation entirely and just build a new function all to do what you want.

MumeClock::parseMSSP(year, month, day, hour)

nschimme (approved):

PR #379, opened 2025-05-10
Add connections between rooms based on movement

ethorondil:
It used to be that moving from room A to room B creates a connection. This seems to have been changed to create connections based on the room IDs given by MUME.  But sometimes MUME does not place room IDs in exits (I think if underground / in the dark).  This commit also adds a connection from room A to room B if you move from A to B, and both A and B have server IDs, during mapping mode.  This makes mapping underground areas possible by just moving back and forth.


nschimme:
@ethorondil Please pull from master. I've made a fix that is hopefully better.

PR #391, opened 2025-07-11
Add character & NPC map-images (in GroupManager and on map)

Sunnyl75:
- Images selected from "Tokens" folder in "Modding" Folder from pop up in GroupManager
- Selections persist.
- Toggle on/off for images in groupmanager and on map
- Resize image in groupmanager with row
- Player image only shown on map if in same room as others.
GroupManager
Multiple players


Sunnyl75:
Superseded by the new Qt6/checkbox defaults PR.

PR #549, opened 2026-09-12
Separate panel logic from widgets and add small-window/touch support

nschimme:
## Summary

Splits the logic that does not need `QWidget` out of several UI classes so it can be unit tested on its own and has a single owner, following the layout `group/`, `timers/` and `adventure/` already use: **domain logic → optional Qt item model → thin view**. The Widgets UI is unchanged in behavior and now sits on top of these pieces. Plus small-window/touch support, a few fixes, and a CI change.

### Map canvas
- **`MapCanvas`** is the rendering/input logic (a `QObject`); **`MapCanvasWindow`** is the `QOpenGLWindow` that hosts it through the **`MapCanvasHost`** interface (repaint scheduling, size, device pixel ratio, cursor).
- **`MapScroller`**: `MapWindow`'s scroll math and continuous-scroll timer.

### Main window
- **`ProxyHost`**: the interface `Proxy`/`ConnectionListener` use to reach the object that owns them; `MainWindow` implements it.
- **`MapLoadHelper`**: map format detection and loading, in `mapstorage/`.

### Models backing the widget panels/dialogs
- `TimerModel` roles (`TimerWidget`), `RoomModel` (`RoomWidget`), `GroupModel` (`GroupWidget`, which also reads its `anyMana` state for the mana column), `AdventureLogModel` (`AdventureWidget` is a `QTextEdit` view of it).
- `UpdateChecker` (`UpdateDialog`), `ClockStrings` (`MumeClockWidget`), `InputHistory`/`TabHistory` (`InputWidget`), `RemoteEditDocumentOps` and `RemoteEditHighlighter` (`RemoteEditWidget`), `StackBlur` (`DescriptionWidget`), `AnsiColorTables` (`AnsiCombo`), `AnsiHtml`'s `AnsiTextToDocument` (`DisplayWidget`).

### Small windows
- `CompactLayout.h` holds the policy (720×480 breakpoint, the client's 55% share of the height).
- A screen below that size starts `MainWindow` in the compact layout; Window → "Compact Layout" toggles it by hand (it does not follow resizes, since flipping the whole window layout under the user is worse than a briefly cramped one). While compact the menu bar is hidden and a row of large buttons in the status bar holds a ☰ button with the top-level menus as submenus (the `QMenu` objects are shared), layer ▲/▼ and center-on-player (operations touch has no wheel or keyboard for; zoom is a pinch); every dock is tabified into one group below the map so one panel shows at a time (the client tab in front), the toolbars are hidden, and the map's scroll bars are suppressed (drag pans); dock title bars are replaced by the tab strip, which uses a larger font so the tabs are finger-sized. The startup choice is judged by the screen (`CompactLayout::probeSize()`), since the expanded layout's minimum size can push the window past a phone's edge, and the window is pulled back onto the screen once the compact minimums apply. The expanded and compact layouts each persist their own window state (`Window State` / `Window State Compact`), so switching never overwrites the other; switching is gated until the first show has restored settings.
- `ClientWidget` reports the configured 80×24 terminal size as its `sizeHint()` rather than its minimum, so the main window can shrink in the first place; the default expanded layout (first run, "Reset Window Layout") still sizes the client dock to it when the window has room for that and the map.
- Kinetic scrolling (`QScroller`) on every panel's scroll area: from touch events when expanded, and from the (synthesized) left button while compact, since a browser can deliver a finger as pointer events rather than touch events; a held press still selects text.

### Touch
- **Map** (all in `MapCanvas`, so any host forwarding events gets them): a long-press opens the context menu (and the finger's release is then ignored, so it does not double as a tap); a flick after a touch drag keeps the view moving with decaying velocity; two fingers pan while pinching; a tap on a room does not raise the hover tooltip.
- When neither the menu bar nor the compact layout's status-bar menu button is on screen, the map's context menu offers a "Menu" submenu holding the top-level menus, so a touch user can always reach them by a long-press (there is no hover to "peek" a hidden menu bar with).
- View → "Center on Player" action (also in the compact button row).
- Menus follow the usual cross-platform layout: **Edit** starts with Undo/Redo and keeps the object submenus (Markers, Rooms, Connections) and Find Rooms; **View** holds the map view (a "Mouse Mode" submenu with every mode, zoom, layers, center, rebuild meshes); a new **Window** menu takes Toolbars, Side Panels and the status bar / scroll bars / menu bar / always-on-top toggles, plus "Compact Layout" and "Reset Window Layout" (the constructor's arrangement for the expanded layout, the default tabbed group for the compact one); **Tools** takes Mode (Play/Map/Offline) and Preferences. Quit, Preferences, About and Check for Update carry `QAction` menu roles so macOS places them in the application menu regardless of wording.
- **Client**: in the compact layout, ↑/↓/⇥ buttons beside the input drive command history and completion; a long-press on the output opens its standard context menu (copy, select all).
- Select Rooms mode on touch: a tap on a room that is not yet selected adds it (a drag from there rubber-bands additively), like Ctrl+click, since touch has no Ctrl; a selected room still moves and empty space still clears. Mouse input is unchanged.

### Input
- Input-method hints on the client input (no auto-capitalization or prediction) and the password fields (additionally `ImhSensitiveData`), so on-screen keyboards behave for MUD commands and credentials.

### Dialogs
- Dialogs whose preferred size does not fit the screen open maximized (`mmqt::showFittedToScreen()`); while compact the Preferences dialog stacks its page list above the pages, can shrink below the pages' natural width (they scroll), and its pages and page list scroll by touch.

### Graphics / wasm memory
- **Render scale** (Graphics page; `Configuration::canvas.renderScale`, default 100%): the map's offscreen framebuffer is rendered at 50/75/100% of the display resolution and presented at full size, cutting its GPU memory for tablets. `Legacy::Functions` distinguishes the render-target pixel ratio (glyph atlases, line widths, the FBO) from the host's ratio (the present viewport). MSAA is off below 100% since it multiplies the same memory.
- wasm: the performance overlay shows the wasm heap and the FBO's footprint.

### Preferences
- General page: **"UI font scale"** spin box for `Configuration::general.uiFontScale` (applied at startup).
- Game Client page (was "Integrated Client"): **How do you play?** — ask each time / play in MMapper / use my own MUD client (`Configuration::general.gameClient`; the web build has only the built-in client).

### First run and the Client Panel
- The Client Panel's welcome page is a card: "Play MUME", "Play music and sound effects", **Play in MMapper**, "I use my own MUD client" with the port to point it at, "Don't ask again" (writes the preference above for the button pressed), and a link to docs.mume.org/newcomers. With the preference set the panel connects at once, or shows "Waiting for your MUD client on localhost, port N" with a "Play here instead" button. Tools → "Play MUME" connects the built-in client; Help gets a "Newcomer's Guide".
- Sound is off for new profiles (volumes default to 0) and is turned on from the card or Preferences → Audio; the "Enable game audio?" bar over the map is gone. On the web build, where the browser needs a user gesture before audio may play, the first key press or tap anywhere unlocks it.
- Client output on touch: a finger scrolls (fed to the viewport's `QScroller` from `DisplayWidget`'s own mouse handlers, since touch arrives as synthesized mouse events) and never selects; the long-press context menu offers Copy / Select All. The peek preview pane is not used while compact.
- Wording: "integrated mud client" is gone from menus and the panel.

### Fixes
- `MapCanvas` takes its `MapCanvasHost` by reference in the constructor (like `FrameManager` takes its window) instead of a QObject parent plus `setHost()`: the parent's `ChildAdded` event reached `MapCanvasWindow::event()` before the member was constructed (UBSan).

### Tests
- New `TestClient` (`InputHistory`/`TabHistory`, `RemoteEditDocumentOps`) and `TestMedia` (`StackBlur`); `TestMainWindow` covers `MapScroller` and the `CompactLayout` policy; `TestCTimers`, `TestGroup`, `TestAdventure`, `TestRoomManager`, and `TestClock` cover the new models and helpers.

### CI
- `build-appimage` installs Qt 6.8.3 via `install-qt-action` (like the AppX/macOS/Windows jobs) instead of the KDE Neon apt repo, whose single host was returning 502s. `external/qtkeychain`'s build-from-source fallback is now allowed on Linux so QtKeychain is built against that Qt.

## Test plan
- [x] Debug build with `CHECK_ODR` + ASan/UBSan: 16/16 tests pass
- [x] RelWithDebInfo build with `WITH_AUDIO=OFF`
- [x] Compact ↔ expanded transitions and per-layout persistence across relaunch (traced on an isolated profile)
- [ ] Manual smoke of the Widgets UI (panels, Preferences, Find Rooms, remote editor, update check); touch long-press on real hardware

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QFncyBYH55KbWPdvzvxXQw


nschimme:
CI note: `build-appimage` is failing in the "Add KDE Neon repository" setup step — `archive.neon.kde.org` is returning HTTP 502 (reproduced on two consecutive heads today, and the host times out from outside CI too). That step runs before any project code is built and this PR doesn't touch the workflow, so it isn't caused by this change; the last `master` run of that workflow succeeded. Re-running the job once; if the mirror is still down it will need another kick once it recovers.

